PR #2, opened 2017-08-29
Create annotations.md

mtwo:
This is a list of standard annotations that Census should support out of the box. This means:

1. Census libraries should attempt to determine appropriate values for these annotations and then attach them to each trace / span
2. Census libraries should use the annotation names supplied in this list rather than creating their own

Custom annotations are of course encouraged, as are vendor-specific ones. This is simply the global list, and thus it applies to all Census clients.

### DIFF
--- a/annotations.md
+++ b/annotations.md
@@ -1,0 +1,23 @@
+Annotation Name      |Description
+---------------------|---
+/rpc/request/size    |
+/rpc/response/size   |
+/http/method         |
+/http/status_code    |
+/grpc/status         |
+/http/url            |
+/http/host           |
+/ip/v4               |
+/ip/v6               |
+/stacktrace          |
+/http/client_city    |
+/http/client_country |
+/http/client_region  |
+/sql/query_string    |
+/agent_name          |
+/agent_version       |
+kube/cluster_name    |
+kube/namespace_id    |
+kube/pod_id          |
+kube/container_name  |
+kube/service         |


adriancole:
I'd suggest saying that it should be possible to extract these OOB (for relevant platforms). For example, it would be common for people to choose to not compute or record certain values. Also some will need a fair amount of summarizing, while others might be self-explanatory.  Finally, I think product-specific things, like kube could be depriortized ex broken after a line break vs portable ones, like request size.

for impl thoughts.. in brave we have a separation between the capability to parse something
https://github.com/openzipkin/brave/blob/master/instrumentation/http/src/main/java/brave/http/HttpAdapter.java
https://github.com/openzipkin/brave/blob/master/instrumentation/http/src/main/java/brave/http/HttpServerAdapter.java

and.. the policy of what you actually choose to record
https://github.com/openzipkin/brave/blob/master/instrumentation/http/src/main/java/brave/http/HttpParser.java

https://github.com/openzipkin/brave/tree/master/instrumentation/http#span-data-policy

food for thought

mtwo:
Good call, I'll break out the kubernetes labels into a vendor / purpose specific group.

Your out of band extraction comment is more related to how Census interacts with these annotations rather than the list of special / significant names, correct?

saturnism:
iirc in zipkin there is a way to associate the service name rather than the url of the service. service name is what's being displayed in zipkin UI. would that be a standard annotation too?

adriancole:
Suppose so, unless we make an endpoint type.

mtwo:
So should we add ServiceName or an equivalent here as well?

adriancole:
It is very used in zipkin, particularly local service name (constant per
tracer) and remote service name (used when you call out to something
uninstrumented like mysql)

bogdandrutu:
Any update on this?

tcolgate:
I'm not sure if I missed it in the spec, but there could be some clearer discussion of annotations vs filters (annotations seem more akin to OpenTracing log entries maybe?)

PR #3, opened 2017-09-08
Make example field values match

duggelz:
Also add all 0's byte to example, to make it clear that single bytes are allowed to be 0.

### DIFF
--- a/encodings/BinaryEncoding.md
+++ b/encodings/BinaryEncoding.md
@@ -44,7 +44,7 @@ were received) or until the first unknown field_id.
 
 ### How can we add new fields?
 If we follow the rules that we always append the new ids at the end of the buffer we can add up 
-to 127. 
+to 127.
 
 TODO(bdrutu): Decide what to do after 127: a) use varint encoding or b) just reserve 255 as a 
 continuation byte.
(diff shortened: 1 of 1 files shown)


duggelz:
I can't assign reviewers, so please review, @bogdandrutu, and CC @liyanhui1228

encodings/BinaryEncoding.md
```diff
@@ -61,15 +61,17 @@ and use the default values.
 * `len` = 16
 
 Is the ID of the whole trace forest. It is represented as a 16-bytes array,
-e.g. (in hex), `4bf92f3577b34da6a3ce929d0e0e4736`. All bytes 0 is considered invalid.
+e.g. (in hex), `4bf92f3577b34da6a3ce929d000e4736`. All bytes 0 is considered invalid.  If converted
+to a 128 bit integer, these bytes are interpreted in big-endian order.
```
bogdandrutu: I would not mention numbers and endianness anywhere and I would try to implement the library using byte[] if possible.

See data model also mention 8-byte array:
https://github.com/census-instrumentation/opencensus-proto/blob/master/trace/span.proto

I think it will be great to not have to deal with endianness. I know that the previous HTTP format used decimal for spanId and probably we will use big-endian there.

What do you think?

adriancole: If we were to mention endian-ness, we could do it in a FAQ or section at the bottom. mainly because it is common practice for better or for worse.

An example could say..

> We encourage IDs to be treated as opaque byte arrays or lower-hex strings. If you need to store them as numbers, be consistent with endian-ness. For example, if you store as big-endian, you'll need to retrieve them as big-endian as well.

duggelz:
I added the word "opaque" and removed all discussion of endianness and int64/int128 interpretations.

encodings/BinaryEncoding.md
```diff
@@ -60,16 +60,16 @@ and use the default values.
 * `field_id` = 0
 * `len` = 16
 
-Is the ID of the whole trace forest. It is represented as a 16-bytes array,
-e.g. (in hex), `4bf92f3577b34da6a3ce929d0e0e4736`. All bytes 0 is considered invalid.
+Is the ID of the whole trace forest. It is represented as an opaque 16-bytes array,
+.g. (in hex), `4bf92f3577b34da6a3ce929d000e4736`. All bytes 0 is considered invalid.
```
bogdandrutu: Removed "e."

duggelz: Fixed.

Issue #4, opened 2017-09-13
Sampling expectations and propagation when not sampled


tsloughter:
There is need for defined points a library or instrumentation should invoke the configured sampler logic and how/what should be propagated to any children of a disabled span. 

When the sampler should be called:

* entering a service (the initial creation of a span or context from a propagated trace context)
* root span creation
* user set a specific sampler for that Span (see SpanBuilder in Java)

In the event the sampler returns `false` and the trace context is propagated to another service, the trace id and parent span id of the span would be propagated. For example if we have services `A`, `B` and `C` with calls resulting in the propagation `A->B->C` and `B` is disabled by its sampler logic then the trace context received by `C` will have the trace id and span id that was originally sent from `A` to `B`.

Note that the fact a parent service who propagated context to a service was or wasn't sampled can be used in the sampling decision of the service, but it having been sampled or not does not create a strict requirement one way or the other (unless of course the configured sampler logic on the service is simply to return the decoded boolean sent with the trace context).

Additionally this defines that once a trace is enabled within a service through its sampler logic, the children within the service of that span do not check with the sampler logic. Meaning within `A`, `B`, or `C` there can not be gaps in the trace.

adriancole:
@bogdandrutu this is actually a big part of what I was mentioning earlier today. we need to define what the semantics are for propagating a "no". Biggest deal is here, where if we provision anyway, we can create gaps which could possibly not be filled.

ex. A ->(yes) B -> (no) C -> (no) D -> (yes) E
if this looks the same as a passthrough (ex we propagating incoming span ID on no), then it looks the same as a trace A -> B -> E (which will have no gaps). However, other consumers of id will be confused (like log entries)

bogdandrutu:
I would call this done. we wrote the document about sampling, if more infos are needed we should open a new issue with concrete action items.

PR #6, opened 2017-09-22
Add details about sampling and minimal details about Span and TraceConfig.

bogdandrutu:


bogdandrutu:
Address the sampling expectation from https://github.com/census-instrumentation/opencensus-specs/issues/4

trace/Sampling.md
```diff
@@ -0,0 +1,32 @@
...
+
+### What kind of samplers does OpenCensus support?
+* `AlwaysSample` - sampler that always makes a "yes" decision every time.
+* `NeverSample` - sampler that always makes a "no" decision every time.
+* `Probability` - sampler that tries to uniformly sample traces with a given probability. When 
+applied to a child `Span` of a **sampled** parent `Span` then it keeps the sampling decision.
+
+### How can users control the Sampler that is used for sampling?
+There are 2 ways to control the `Sampler` used when the library does sampling:
+* Controlling the global default `Sampler` via [TraceConfig](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/TraceConfig.md).
```
tsloughter: To be clear. Global refers to within the service being traced at that point in time, right? Not some sort of global configuration for all services within the scope of the distributed trace.

trace/Sampling.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Sampling Bit (propagated via TraceOptions)
+
+Sampling bit is always set only at the start of a Span, and is using a `Sampler`
+
+### What kind of samplers does OpenCensus support?
+* `AlwaysSample` - sampler that always makes a "yes" decision every time.
+* `NeverSample` - sampler that always makes a "no" decision every time.
+* `Probability` - sampler that tries to uniformly sample traces with a given probability. When 
+applied to a child `Span` of a **sampled** parent `Span` then it keeps the sampling decision.
```
tylerbenson: What about a fixed rate sampling? This might bake extrapolation more complicated, but seems useful to maintain a steady flow of traces regardless of traffic flow.

trace/Span.md
```diff
@@ -0,0 +1,32 @@
+# Span API
+
+This class represents a Span. A span contains a SpanContext and  allows users to record 
+tracing events based on the data model defined [here](https://github.com/census-instrumentation/opencensus-proto/blob/master/trace/trace.proto).
+All Span implementations MUST not allow users to read the recorded data,
```
tylerbenson: This is an interesting requirement. Why is this important?

adriancole: Probably this is about the span api, as there've been some things in the past where there's readback race conditions which aren't worth explaining. Brave used to have synchronized readback, we removed it, and will likely only support this as a pre-report hook. Not sure bogdan's rationale though!

trace/Sampling.md
adriancole: suppose this is a matter of required vs optional? I think it is a sensible option

trace/Sampling.md
adriancole: yep

trace/Sampling.md
```diff
@@ -0,0 +1,32 @@
...
+
+### How can users control the Sampler that is used for sampling?
+There are 2 ways to control the `Sampler` used when the library does sampling:
+* Controlling the global default `Sampler` via [TraceConfig](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/TraceConfig.md).
+* Pass a specific `Sampler` when start the [Span](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/Span.md)
```
adriancole: add a sub-bullet explaining that AlwaysSampler and NeverSampler can be used to implement request-specific decisions such as those based on http paths. It isn't intuitive (although intuitive afterwards!)

trace/Sampling.md
```diff
@@ -0,0 +1,32 @@
...
+
+### When does OpenCensus sample traces?
+The OpenCensus library does sampling based on the following rules:
+1. If the span is a root `Span` then a `Sampler` will be used to make the sampling decision:
+   * If user provided a `Sampler`, use it to determine the sampling decision.
```
adriancole: user-provided, might be better put as "span-scoped" ex the global sampler might also be configured by the user and the span-scoped might not be (as it is employed by a layered http sampler)

trace/Sampling.md
```diff
@@ -0,0 +1,32 @@
...
+   * Else use the global default `Sampler` to determine the sampling decision.
+2. If the span is a child of a remote `Span` the sampling decision will be:
```
adriancole: we should mention how someone can retain the remote decision as this is common practice in internal network tracing (and all B3 for example)

trace/Sampling.md
```diff
@@ -0,0 +1,32 @@
...
+This document is about the sampling bit, sampling decision, samplers and how and when does 
+OpenCensus sample traces. A sampled traces is a traces that gets exported via the configured
+exporters.
+
+## Sampling Bit (propagated via TraceOptions)
+
+Sampling bit is always set only at the start of a Span, and is using a `Sampler`
+
+### What kind of samplers does OpenCensus support?
+* `AlwaysSample` - sampler that always makes a "yes" decision every time.
```
adriancole: PropagateDecision? it seems we don't so maybe we need a sampler that does?

trace/Sampling.md
Licenser: I was wondering about something along the line too, is this a list of required samplers or is this an exhaustive list of allowed samplers?

adriancole:
>
> I was wondering about something along the line too, is this a list of
> required samplers or is this an exhaustive list of allowed samplers?
>
suspect it can only be the minimal, as we can't prevent third party. maybe
we can change the wording a bit

trace/Span.md
bogdandrutu: So I would like the OpenCensus libraries to have the same "look and feel". I think it is hard to get it right in all the languages (if we really need it we can discuss and add it later but I need to think more about what is really needed). Removing an API is much harder than adding a new one.

The rational behind this is the following:
* I would like to achieve the same "look and feel" for all the languages.
* If you allow  any getter then that means the data needs to be stored in the Span object all the time and they cannot be streamed directly to the repository for example.
* If you have to store all the data because you want to provide getters then you for refcounting (in languages like C++ where no GC).
* If you force refcounting you force heap-allocation, which in some C++ systems may be expensive.

I think for the moment adding getters in some languages limits the implementation design and I would prefer to not have it until we have a strong argument to support that.

trace/Sampling.md
bogdandrutu: So I see AlwaysSample and NeverSample just as samplers to control specific requests/paths. Usually as default samplers we will use something like the ProbabilitySampler which will propagate the decision.

PropagateDecision only can be achieved with a ProbabilitySampler(0.0).

trace/Sampling.md
bogdandrutu: I don't want to increase the requirements from the beginning. I agree fixed rate sampling adds a great value. Can you please add an issue to discuss this? Probably we should prototype a reference implementation for that.

trace/Sampling.md

trace/Sampling.md

trace/Sampling.md
bogdandrutu: Not sure what you mean here? I think I mentioned that ProbabilitySampler retain the sampling decision (AlwaysSampler same). NeverSample does not retain is a kind of force "unsampled".

bogdandrutu:
This is a list of required samplers that we should implement in all the languages. We can add more later, and we also allow users to implement their own.

bogdandrutu:
PTAL. Let me know if I did not address all the issues.

trace/Sampling.md
```diff
@@ -0,0 +1,35 @@
+# Sampling
+
+This document is about the sampling bit, sampling decision, samplers and how and when does 
+OpenCensus sample traces. A sampled traces is a traces that gets exported via the configured
+exporters.
+
+## Sampling Bit (propagated via TraceOptions)
+
+Sampling bit is always set only at the start of a Span, and is using a `Sampler`
```
adriancole: I think we need an overview of propagation (vs a state chart). or at least a separate section on how to achieve B3 style "trust the caller" decisions. The fact that a sampler is required for each span may seem complex to those who currently only do it at the head of a trace. having a default as create or trust. would match zipkin and zipkin-like systems.. worth either making default do that or making it easy to have the default do that imho. I don't think probability 0 is intuitive as for ex in zipkin this is same as Never

bogdandrutu: so the default is Probability(1.0/10000):
* As Probability sampler it keeps the sampling probability;
* It inflates the sampling rate to 1/10000 (in case no-sampled);

adriancole:
> so the default is Probability(1.0/10000):
>
> As Probability sampler it keeps the sampling probability;
> It inflates the sampling rate to 1/10000 (in case no-sampled);

I think a PreferRemote decorator should be present to allow trusting
the upstream decision (ex a no decision). This is the default in
sleuth for example, which doesn't ever start new trace but always
resumes them (nice for when a proxy is in front).

I agree that re-assessing decisions is helpful, for example,
@mclarke47 mentioned that one issue with envoy is that it doesn't have
a path-specific sampling policy, which flooded their zipkin server
with useless traces (think favicon.ico). IOTW, I think it is very
valid to distrust the upstream decision, but at least in zipkin the
majority do trust upstream. Having a way to express that makes it
easier for folks to transition.

trace/Sampling.md
tylerbenson: https://github.com/census-instrumentation/opencensus-specs/issues/9

PR #8, opened 2017-09-28
Start documenting the binary encoding for Tag Context.

sebright:
/cc @bogdandrutu @AlekseySanin @dinooliva @songy23 @adriancole

### DIFF
--- a/encodings/BinaryEncoding.md
+++ b/encodings/BinaryEncoding.md
@@ -100,6 +100,46 @@ This corresponds to:
 * `spanId` = {52, 240, 103, 170, 11, 169, 2, 183}
 * `traceOptions` = 1
 
+### Tag Context
+The Tag Context format uses Varint encoding, which is described in
+https://developers.google.com/protocol-buffers/docs/encoding#varints. Note that
+each type of tag can appear multiple times.
+
+#### String tag
+
+* `field_id` = 0
+* `field_format` = `<tag_key_len><tag_key><tag_val_len><tag_val>` where
+
+  * `tag_key_len` is a varint encoded integer.
+  * `tag_key` is `tag_key_len` bytes comprising the tag key name.
+  * `tag_val_len` is a varint encoded integer.
+  * `tag_val` is `tag_val_len` bytes comprising the tag value.
+
+#### Integer tag
+
+* `field_id` = 1
+* `field_format` = `<tag_key_len><tag_key><tag_val>` where
+
+  * `tag_key_len` is a varint encoded integer.
+  * `tag_key` is `tag_key_len` bytes comprising the tag key name.
+  * `tag_val` is 8 bytes, a little-endian int64, representing the tag value.
+
+#### True boolean tag
+
+* `field_id` = 2
+* `field_format` = `<tag_key_len><tag_key>` where
+
+  * `tag_key_len` is a varint encoded integer.
+  * `tag_key` is `tag_key_len` bytes comprising the tag key name.
+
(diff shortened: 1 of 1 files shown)


encodings/BinaryEncoding.md
```diff
@@ -100,6 +100,46 @@ This corresponds to:
...
+
+* `field_id` = 0
+* `field_format` = `<tag_key_len><tag_key><tag_val_len><tag_val>` where
+
+  * `tag_key_len` is a varint encoded integer.
+  * `tag_key` is `tag_key_len` bytes comprising the tag key name.
+  * `tag_val_len` is a varint encoded integer.
+  * `tag_val` is `tag_val_len` bytes comprising the tag value.
+
+#### Integer tag
```
bogdandrutu: Long tag?

adriancole: I think integer is fine. it is used in other specs (less confusing in a way to start with that it is an integer, then drill into properties like how long it is) https://tools.ietf.org/html/rfc7541#section-5.1

encodings/BinaryEncoding.md
```diff
@@ -100,6 +100,46 @@ This corresponds to:
...
+
+* `field_id` = 1
+* `field_format` = `<tag_key_len><tag_key><tag_val>` where
+
+  * `tag_key_len` is a varint encoded integer.
+  * `tag_key` is `tag_key_len` bytes comprising the tag key name.
+  * `tag_val` is 8 bytes, a little-endian int64, representing the tag value.
+
+#### True boolean tag
```
adriancole: I would merge these under Boolean tag, then just say the field ID is the value, ex 2 is true and 3 is false

encodings/BinaryEncoding.md
adriancole: unless.. we expect to have separate integer types (which I'm personally not a fan of). then long or i64 is appropriate in the heading

encodings/BinaryEncoding.md
sebright: Done

encodings/BinaryEncoding.md
sebright: I think integer is clearer while we only have one integer type.  Could we just change the heading if/when we add a new type of integer?

songy23: I would suggest for Int64 since we mentioned  _`tag_val` is 8 bytes, a little-endian int64, representing the tag value_. Though `Integer` is also fine since we don't have other int types.

Also refer to https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.metricDescriptors#ValueType

sebright: > I would suggest for Int64 since we mentioned tag_val is 8 bytes, a little-endian int64, representing the tag value. Though Integer is also fine since we don't have other int types.

I'll ask for another opinion, since we are split on the choice of heading. /cc @maxim-l 

> Also refer to https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.metricDescriptors#ValueType

This section is just for the encoding, so I think that we should have a separate file that describes the in-memory tag context and the choice of tag value types.

maxim-l: As a .md section title, I'm with "Integer" wording: it aligns with informal string and bool wording and there's info below that it's 8 bytes and int64 is mentioned.

sebright: Thanks.  I'll leave it as "integer" and merge once the CLA check passes.

PR #17, opened 2017-10-10
Add more details to the general binary encoding specification (closes #11).

### LINKED ISSUE #11: Questions about the binary encoding
I would like to make a pull request to clarify some parts of the [binary encoding specification](https://github.com/census-instrumentation/opencensus-specs/blob/master/encodings/BinaryEncoding.md).

1. Do fields always need to be serialized in order?  I.e., in version 1, do all fields with id 0 need to precede all fields with id 1, even if both fields were known when version 1 was first released?
2. Can a field disallow multiple values?  For example, when a serialized trace context contains multiple trace ids, could that be considered a parse error?
3. Is the version mentioned in [Serialization Rules](https://github.com/census-instrumentation/opencensus-specs/blob/master/encodings/BinaryEncoding.md#serialization-rules) a different version from the version described at the start of [General Format](https://github.com/census-instrumentation/opencensus-specs/blob/master/encodings/BinaryEncoding.md#general-format)?  If I understand correctly, the `version` field describes something like a major version while the version mentioned in Serialization Rules is like a minor version.

/cc @bogdandrutu

sebright:
This commit adds details about field order, repeated fields, and data type
specific versions.

### DIFF
--- a/encodings/BinaryEncoding.md
+++ b/encodings/BinaryEncoding.md
@@ -1,5 +1,9 @@
 # BINARY FORMAT
 
+The binary format can be used to encode different data types, each with different fields. This
+document first describes the general format and then applies it to specific data types,
+including Trace Context and Tag Context.
+
 ## General Format
 Each encoding will have a 1 byte version followed by the version format encoding:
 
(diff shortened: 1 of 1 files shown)


encodings/BinaryEncoding.md
```diff
@@ -10,37 +14,45 @@ This will allow us to, in 1 deprecation cycle to completely switch to a new form
...
 
 * `field_id` is a single byte.
 
-* `field_format` must be defined for each metadata field separately, that means that for field_id
- = 0 in trace context the field_value may have a completely different representation than the 
- field_id = 0 in the server-stats metadata.
+* `field_format` must be defined for each field separately.
+
+The specification for a data type's format must also specify whether each field is single or
+repeated. For example, String tag in Tag Context is repeated, and Trace-id in Trace Context in
```
songy23: The last "in" should be "is".

encodings/BinaryEncoding.md
```diff
@@ -10,37 +14,45 @@ This will allow us to, in 1 deprecation cycle to completely switch to a new form
...
+single. Every single field is optional. The specification for a data type's format MAY define a
```
songy23: > Every single field is optional.

I feel a little ambiguous here. Does it mean "all the fields are optional" or "only single fields are optional, and repeated fields are required"?

encodings/BinaryEncoding.md
```diff
@@ -10,37 +14,45 @@ This will allow us to, in 1 deprecation cycle to completely switch to a new form
...
+default value for any single field, which must be used when the field is missing.
```
songy23: > for any single field

Same here.

encodings/BinaryEncoding.md
sebright: Done

encodings/BinaryEncoding.md
sebright: I meant the second one.  I changed "single" to "optional", as in Protocol Buffers, to try to make it clearer.

sebright:
Thanks!

PR #20, opened 2017-10-19
Add details about Span creation and context interaction.

bogdandrutu:


trace/Span.md
```diff
@@ -24,9 +27,54 @@ parent span id for the child Span.
...
+When an explicit propagated Context is used implementation MUST create a new Context when a Span 
+is attached (immutable Context):
+* When attach/detach an already created Span the API MAY be called `WithSpan`.
+* When attach/detach at the creation time the API MAY be called `StartSpan` or `StartScopedSpan`.
+
+### What is Span Lifetime?
+Span lifetime represents the process of recording the start and the end timestamps to the Span 
+object:
+* The start time is recorded when the Span is created.
+* The end time needs to be recorded when the operation is ended.
```
adriancole: Add clarification that startScopedSpan does imply closing the scope finishes the lifetime. Not sure it is made obvious yet

trace/Span.md
```diff
@@ -1,8 +1,11 @@
-# Span API
+# Span
 
-This class represents a Span. A span contains a SpanContext and  allows users to record 
-tracing events based on the data model defined [here](https://github.com/census-instrumentation/opencensus-proto/blob/master/trace/trace.proto).
-All Span implementations MUST not allow users to read the recorded data, 
+Span represents a single operation within a trace. Spans can be nested to form a trace tree.
```
duggelz: We talk about a "trace" here, but it's not given a definition, and more importantly, there's no class or object in the code that directly maps to a "trace".  Can you add a "Trace" heading above "Span" (or maybe in the README?) that says this?

trace/Span.md
```diff
@@ -24,9 +27,54 @@ parent span id for the child Span.
...
+Context interaction represents the process of attaching/detaching a Span to the Context 
+in order to propagate it in-process (possibly between threads) and between function calls.
+
+There are two supported implementations for the Context based on how the propagation is implemented:
+* With implicit propagation - implicitly passed between function calls and threads, usually 
+implemented using thread-local variables (e.g. Java [io.grpc.Context][javaContext])
+* With explicit propagation - explicitly passed between function calls and threads (e.g. Go 
+[Context][goContext])
+
+When an implicit propagated Context is used implementation MUST use scoped objects to
```
duggelz: used implementation -> used, the implementation

README.md
```diff
@@ -65,7 +65,7 @@ The key elements of the Context Support are:
 that allows different sub-contexts to be propagated.
 * Languages that already have this support, like [Go][goContext] or C# (ExecutionContext), MUST 
 use the language supported generic context instead of building their own.
-* For an explicit generic context implementation you can look at the Java [io.grpc.Context][gRPCContext].
+* For an implicit generic context implementation you can look at the Java [io.grpc.Context][gRPCContext].
```
rakyll: What does implicit mean in this context. I used to think io.grpc.Context is explicit because the user creates and uses it to explicitly scope a block run in that specific context. Maybe it is too early here to mention implicit/explicit.

trace/Span.md
```diff
@@ -1,8 +1,11 @@
-# Span API
+# Span
 
-This class represents a Span. A span contains a SpanContext and  allows users to record 
-tracing events based on the data model defined [here](https://github.com/census-instrumentation/opencensus-proto/blob/master/trace/trace.proto).
-All Span implementations MUST not allow users to read the recorded data, 
+Span represents a single operation within a trace. Spans can be nested to form a trace tree. 
+Often, a trace contains a root span that describes the end-to-end latency and, optionally, one or
```
rakyll: This is hard to follow. You may explain traces are a tree. There is a root span and many nodes that represents worked started from the root point.

trace/Span.md
```diff
@@ -1,8 +1,11 @@
...
+more sub-spans for its sub-operations.
+ 
+A span contains a SpanContext and allows users to record tracing events based on the data model
```
rakyll: Isn't span context also the identifier. We should focus on that responsibility before mentioning anything about recording tracing events.

trace/Span.md
```diff
@@ -24,9 +27,54 @@ parent span id for the child Span.
 #### TraceOptions 
 Represents the options for a trace. It is represented as 1 byte (bitmap).
 
-##### Supported bits
+##### Supported Bits
```
rakyll: Mixed style.

Either capitalize all the words in the title or revert this line.

trace/Span.md
```diff
@@ -24,9 +27,54 @@ parent span id for the child Span.
...
 * Sampling bit -  Bit to represent whether trace is sampled or not (mask `0x1`).
 
-## Span creation API
+## Span Creation
```
rakyll: ditto

README.md
tsloughter: Yea, this needs more explanation then. I read it and having no knowledge of io.grpc.Context I figured since it was implicit this meant it was some variable automatically tied to a thread by the framework/library.

tsloughter:
There is no mention of what Context is before detailing how a Span interacts with a context. It makes the document read like this use of Context is referring to a SpanContext, when it is talking about the generic Context objects like `io/grpc/Context.java`. I think there should be some mention of the distinction before detailing the interaction between a span and the context.

trace/Span.md
duggelz: It would be good to clarify that SpanContext and Context are completely differently and pretty much unrelated, and SpanContext is really more like SpanHandle or SpanName or SpanControlBlock or SpanIdentifier (except span_id is already used).  Ok, I can't think of a good name either.

trace/README.md
```diff
@@ -1,10 +1,14 @@
...
-* [TraceConfig API]()
+## Main Components
+* [Span][SpanLink]
+* [TraceConfig][TraceConfigLink]
 
 ## Utils
-* [Sampling logic](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/Sampling.md)
+* [Sampling logic][SamplingLink]
+
+[SpanLink]: https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/Sampling.md
```
sebright: Should this be Span.md?

trace/Span.md
```diff
@@ -24,9 +27,54 @@ parent span id for the child Span.
...
+* When attach/detach an already created Span the API MAY be called `WithSpan`.
+* When attach/detach at the creation time the API MAY be called `StartSpan` or `StartScopedSpan`.
+
+### What is Span Lifetime?
+Span lifetime represents the process of recording the start and the end timestamps to the Span 
+object:
+* The start time is recorded when the Span is created.
+* The end time needs to be recorded when the operation is ended.
+
+### Why to support Spans are not always attached to the Context?
```
sebright: I think this is clearer: "Why support Spans that are not attached to the Context?"

trace/Span.md
```diff
@@ -24,9 +27,54 @@ parent span id for the child Span.
...
-## Span creation API
+## Span Creation
+The implementation MUST allow users to create two types of Spans:
+* Root Spans - spans that do not have a parent.
+* Child Spans - the parent can be explicitly set or inherit from the Context.
 
-TODO(bdrutu): Add details here.
\ No newline at end of file
+When creating a Span the implementation MUST allow users to create the Span attached or detached 
+from the Context, this allows users to manage the interaction with the Context independently of
```
sebright: > Span attached or detached from the Context

I think "detached from the Context" is a little unclear if this case also refers to creating a span without a context, since "detached" could imply that there was a context.

trace/Span.md
```diff
@@ -24,9 +27,54 @@ parent span id for the child Span.
...
+### What is Span Lifetime?
+Span lifetime represents the process of recording the start and the end timestamps to the Span 
+object:
+* The start time is recorded when the Span is created.
+* The end time needs to be recorded when the operation is ended.
+
+### Why to support Spans are not always attached to the Context?
+* Allow users to use the OpenCensus library without using a Context.
+* Allow users to have more control for the lifetime of the Span.
+* There are cases for example HTTP/RPC interceptors where the Span creation and usage happens in
```
sebright: It would help to add commas, such as "There are cases, for example HTTP/RPC interceptors, where the Span creation and usage happens in".

trace/Span.md
bogdandrutu: done.

trace/Span.md
bogdandrutu: done.

trace/README.md
bogdandrutu: done.

trace/Span.md
bogdandrutu: done.

trace/Span.md
bogdandrutu: done.

trace/Span.md
bogdandrutu: done

bogdandrutu:
Merging this now and I will address the rest of the comments in a different PR.

PR #23, opened 2017-10-24
Add note that namespace and pacakges are suggestions not hard rules

rakyll:


NamespaceAndPackage.md
```diff
@@ -12,7 +14,7 @@ The second level package names MUST be:
 * **context**: for all context related functionality
 * **trace**: for all trace functionality
 * **common**: for all public API components that are shared across multiple of the above. These are
-typically utility classes such as timestamps, etc.
+typically utility classes such as timestamps, etc. For Go, use the top-level package instead.
```
bogdandrutu: You should change MUST -> SHOULD then (5 lines above)

rakyll: Done

PR #25, opened 2017-11-08
Specify first version and optional/repeated for each field in BinaryEncoding.md.

sebright:
The version number is important for serializing field values in version order.

### DIFF
--- a/encodings/BinaryEncoding.md
+++ b/encodings/BinaryEncoding.md
@@ -69,24 +69,29 @@ and use the default values.
 
 ### Trace Context
 
-#### Trace-id
+#### Fields added in Trace Context version 0
 
+##### Trace-id
+
+* optional
 * `field_id` = 0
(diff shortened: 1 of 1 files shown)


sebright:
ping

encodings/BinaryEncoding.md
```diff
@@ -69,24 +69,29 @@ and use the default values.
 
 ### Trace Context
 
-#### Trace-id
+#### Fields added in Trace Context version 0
```
rakyll: What about removing "Fields added in" and just calling it "Trace Context version 0"?

sebright: I wanted to show the full history of fields in this file, because it affects the serialization order.  Fields can be serialized in any order if they were added in the same version, but all fields added in an earlier version must precede all fields added in a later version.

PR #28, opened 2017-12-05
Tweak the trace sampling decision docs

odeke-em:


odeke-em:
Thank you for the review @sebright!

PR #37, opened 2018-01-10
WIP: HTTP tracing

ramonza:


odeke-em:
@ramonza thanks for the PR, weirdly googlebot can't find your email though.

trace/HTTP.md
```diff
@@ -0,0 +1,25 @@
+# Tracing HTTP
+
+## Client
+
+* Client span name format
+* Message send events
+* Message receive events
+* Status
+* Attributes names
+** Host
```
rakyll: We might need query parameters here as well.

trace/HTTP.md
```diff
@@ -0,0 +1,25 @@
...
+** Resolved IP of connected socket
+** Path
+** Protocol version
+* Transport layer details (how do we represent TLS negotiation)
+* Redirects (how are these to be represented)
```
rakyll: Shouldn't each redirect represented as its own span?

trace/HTTP.md
```diff
@@ -0,0 +1,25 @@
...
+* Client span name format
+* Message send events
+* Message receive events
+* Status
+* Attributes names
+** Host
+** Resolved IP of connected socket
+** Path
+** Protocol version
+* Transport layer details (how do we represent TLS negotiation)
```
rakyll: We can add more such as DNS and more.

trace/HTTP.md
```diff
@@ -0,0 +1,25 @@
...
+* Redirects (how are these to be represented)
+* Retries (should be logged as separate requests)
+* https://blog.golang.org/http-tracing
```
rakyll: This might be under transport layer details.

trace/HTTP.md
adriancole: yes

trace/HTTP.md
```diff
@@ -0,0 +1,25 @@
+# Tracing HTTP
+
+## Client
+
+* Client span name format
```
adriancole: in zipkin we use the http method name due to fixed cardinality. For platforms that support route templates, that's the better choice. https://github.com/openzipkin/zipkin/issues/1874

PR #41, opened 2018-01-19
Initial doc on HTTP stats detailing just client metrics.

ramonza:


http/stats.md
```diff
@@ -0,0 +1,54 @@
...
+| headers_recv       | none                     | Total number of header lines received in responses, not including trailing headers                                                                                                                                                                                                                |
+| latency            | distribution             | Time between first byte of request headers sent to last byte of response received, or terminal error                                                                                                                                                                                              |
+| ttfb               | distribution             | Time between last byte of request sent and first byte of response body received                                                                                                                                                                                                                   |
```
bogdandrutu: This is very hard to measure correctly, I would remove it or put an optional for this.

http/stats.md
```diff
@@ -0,0 +1,54 @@
...
+
+### Measures
+
+Client stats are recorded for each individual HTTP request, including for each redirect (followed or not). 
+
+| Measure suffix     | Default View Aggregation | Description                                                                                                                                                                                                                                                                                       |
+|--------------------|--------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| started            | sum                      | Number of all client requests started                                                                                                                                                                                                                                                             |
```
bogdandrutu: I would not add this for the moment, or maybe just mark it as optional.

http/stats.md
```diff
@@ -0,0 +1,54 @@
...
+| latency            | distribution             | Time between first byte of request headers sent to last byte of response received, or terminal error                                                                                                                                                                                              |
+| ttfb               | distribution             | Time between last byte of request sent and first byte of response body received                                                                                                                                                                                                                   |
+| connections_opened | count                    | Number of underlying transport (TCP/TLS) connections opened                                                                                                                                                                                                                                       |
```
bogdandrutu: Is this the total number of connection or is the active connections?

http/stats.md
ramonza: The way I see it, everything is optional. I meant this as more like if the language provides this metric then this is what it's name should be. Curious why you think this is not an important metric?

http/stats.md
ramonza: Go has a good way to measure this (httptrace) and often it's a useful metric. I will comment at the top of the file that not all libraries provide all stats.

http/stats.md
ramonza: I imagined it was just the total number opened. I can see the latter being useful as well so I'll add a connections_closed as well.

bogdandrutu: Sounds good.

http/stats.md
bogdandrutu: Please comment.

http/stats.md
bogdandrutu: I did not want to say that it is not important, in general we get this information from the tracing side by looking at the number of active HTTP spans.

Happy to keep it as optional (as you mentioned all the metrics are optional).

FYI: also in general for the number of completed requests we get that from the latency distribution by querying the total count you can mentioned that in this doc.

http/stats.md
ramonza: Actually after looking into this further, I think I agree with you that it's not that useful and has definitional problems. I will remove it.

http/stats.md
ramonza: Yes, this is how the view "ended" is defined below.

PR #48, opened 2018-02-01
Update HTTP metrics and views.

ramonza:
* Simplify and reduce requirements
* Bring into closer alignment with gRPC
* Specify units

### DIFF
--- a/http/stats.md
+++ b/http/stats.md
@@ -8,6 +8,8 @@ There is no special support for multi-part HTTP requests and responses. These ar
 
 ## Units
 
+Units are encoded according to the case-sensitive abbreviations from the [Unified Code for Units of Measure](http://unitsofmeasure.org/ucum.html):
+
 * Latencies are measures in float64 milliseconds, denoted "ms"
 * Sizes are measured in bytes, denoted "By"
 * Dimensionless values have unit "1"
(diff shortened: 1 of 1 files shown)


http/stats.md
```diff
@@ -6,24 +6,32 @@ All times (latencies) are measured in float64 fractions of microseconds.
 
 There is no special support for multi-part HTTP requests and responses. These are just treated as one request.
 
+## Units
+
+* Latencies are measures in float64 milliseconds, denoted "ms"
+* Sizes are measured in bytes, denoted "By"
```
HailongWen: I think "By" might be confused with "by". How about using only "B" to denote "byte"?

ramonza: These are from http://unitsofmeasure.org/ucum.html (also used by Stackdriver)

ramonza: Added a note to this effect

HailongWen: I see, thanks!

http/stats.md
```diff
@@ -1,53 +1,60 @@
...
-| status_code   | HTTP status code, or 0 if no response status line was received |
-| response_type | Response media type, if applicable                             |
-| request_type  | Request media type, if applicable                              |
-| version       | HTTP version in request                                        |
-| path          | URL path (not including query string)                          |
-| host          | Value of the request Host header                               |
+| Tag suffix  | Description                                                          |
+|-------------|----------------------------------------------------------------------|
+| http.method | HTTP method, capitalized (i.e. GET, POST, PUT, DELETE, etc.)         |
+| http.status | HTTP status code, or "error" if no response status line was received |
```
odeke-em: Did we mean `http.status_code` here?

PR #49, opened 2018-02-09
Add HTTP Attributes spec

rakyll:
This document summarizes the attributes to be collected on
the HTTP spans.

### DIFF
--- a/trace/HTTPAttributes.md
+++ b/trace/HTTPAttributes.md
@@ -13,21 +13,27 @@ All attributes are optional.
 |---------------------------|-----------------------------|---------------------------------|
 | "http.host"               | Request URL host            | "example.com"                   |
 | "http.method"             | Request URL method          | "GET"                           |
-| "http.path"               | Request URL path            | "/users"                        |
+| "http.path"               | Request URL path            | "/users/25f4c31d"               |
+| "http.route"              | Matched request URL route   | "/users/:userID"                |
 | "http.user_agent"         | Request user-agent          | "HTTPClient/1.2"                |
 | "http.status_code"        | Response status code        | 200                             |
-| "http.request_body_size"  | Request body size in bytes  | 64328                           |
-| "http.response_body_size" | Response body size in bytes | 128                             |
+
+Exporters should always export the collected attributes.
+Exporters should map the collected attributes to backend's
+known attributes/labels.
 
 The following table summarizes how OpenCensus attributes maps to the
-known attributes/labels on tracing backends.
+known attributes/labels on supported tracing backends.
 
-| OpenCensus attribute      | Stackdriver Trace label                     |
-|---------------------------|---------------------------------------------|
-| "http.host"               | "trace.cloud.google.com/http/host"          |
-| "http.method"             | "trace.cloud.google.com/http/method"        |
-| "http.path"               |                                             |
-| "http.user_agent"         | "trace.cloud.google.com/http/user_agent"    |
-| "http.status_code"        | "trace.cloud.google.com/http/status_code"   |
-| "http.request_body_size"  | "trace.cloud.google.com/http/request/size"  |
-| "http.response_body_size" | "trace.cloud.google.com/http/response/size" |
+| OpenCensus attribute      | Zipkin             | Jaeger             | Stackdriver Trace label                    |
+|---------------------------|--------------------|--------------------|--------------------------------------------|
+| "http.host"               | "http.host"        |                    | "trace.cloud.google.com/http/host"         |
+| "http.method"             | "http.method"      | "http.method"      | "trace.cloud.google.com/http/method"       |
+| "http.path"               | "http.path"        |                    |                                            |
+| "http.route"              | "http.route"       |                    |                                            |
+| "http.user_agent"         |                    |                    | "trace.cloud.google.com/http/user_agent"   |
+| "http.status_code"        | "http.status_code" | "http.status_code" | "trace.cloud.google.com/http/status_code"  |
+
+Request body and response size of incoming and outgoing requests should be
+represented as message events. Each redirect should be represented as a
+message event.
--- a/trace/HTTPAttributes.md
+++ b/trace/HTTPAttributes.md
@@ -7,6 +7,8 @@ HTTP integrations should set the following attributes on the client
 and server spans. For a server, request represents the incoming request.
 For a client, request represents the outgoing request.
 
+All attributes are optional.
+
 | Attribute name       | Description            | Example value                   |
 |----------------------|------------------------|---------------------------------|
 | "http.host"          | Request URL host       | "example.com"                   |
--- a/trace/HTTPAttributes.md
+++ b/trace/HTTPAttributes.md
@@ -1,0 +1,35 @@
+# HTTP Attributes
+
+This document summarizes HTTP attributes recorded from HTTP
+requests and responses.
+
+HTTP integrations should set the following attributes on the client
+and server spans. For a server, request represents the incoming request.
+For a client, request represents the outgoing request.
+
+```
+| Attribute name       | Description            | Example value                   |
+|----------------------|------------------------|---------------------------------|
+| "http.url"           | Request URL            | "https://example.com/users?q=a" |
+| "http.host"          | Request URL host       | "example.com"                   |
+| "http.method"        | Request URL method     | "GET"                           |
+| "http.path"          | Request URL path       | "/users"                        |
+| "http.user_agent"    | Request user-agent     | "HTTPClient/1.2"                |
+| "http.status"        | Response status code   | 200                             |
+| "http.request_size"  | Request size in bytes  | 64328                           |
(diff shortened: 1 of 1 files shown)


trace/HTTPAttributes.md
```diff
@@ -0,0 +1,35 @@
...
+and server spans. For a server, request represents the incoming request.
+For a client, request represents the outgoing request.
+
+```
+| Attribute name       | Description            | Example value                   |
+|----------------------|------------------------|---------------------------------|
+| "http.url"           | Request URL            | "https://example.com/users?q=a" |
+| "http.host"          | Request URL host       | "example.com"                   |
+| "http.method"        | Request URL method     | "GET"                           |
+| "http.path"          | Request URL path       | "/users"                        |
```
yurishkuro: host and path are already available from the url, what's the rationale for capturing them again? Tracing backends can easily extract them later.

tsloughter: Or have only host and path and leave out url.

adriancole:
Url also includes query parameters which to some is a security concern but
not others. Also some may choose to use key-based policy as opposed to
deriving data (easy logic less parsing). Defining what constants are is
separate (even if currently coupled) from collection policy. Usually these
constants apply to all languages so can help manage places where something
is cheaper.

Ex Url is usually more expensive to assemble, than path. When one does
collect the url, they can choose to not collect the path even if that isnt
implemented here at the moment.

Tldr I personally dont think we should make it impossible to define keys
for these aspects independently. They are in use today in many tracing
implementations so intentionally deleting them seems to not be great for
folks who want to transition and keep the fidelity they have now.

tsloughter:
Right, the web server I've instrumented does not currently include the url in the request record, and I'd prefer not to have to either add it to the record or assemble in the middleware.

trace/HTTPAttributes.md
rakyll: Due to the security concerns, we should prefer having host and path instead of requiring URL to be recorded. I removed the URL attribute and also marked all attributes as optional. PTAL.

trace/HTTPAttributes.md
```diff
@@ -0,0 +1,33 @@
...
+known attributes/labels on tracing backends.
+
+| OpenCensus attribute | Stackdriver Trace label                     |
+|----------------------|---------------------------------------------|
+| "http.host"          | "trace.cloud.google.com/http/host"          |
+| "http.method"        | "trace.cloud.google.com/http/method"        |
+| "http.path"          |                                             |
+| "http.user_agent"    | "trace.cloud.google.com/http/user_agent"    |
+| "http.status"        | "trace.cloud.google.com/http/status_code"   |
+| "http.request_size"  | "trace.cloud.google.com/http/request/size"  |
```
HailongWen: Maybe a dumb question, is this size the content size (without url,  headers, etc. ) or transport layer message size (http body size) ?

rakyll: They represent the body size, I renamed them and updated the description.

trace/HTTPAttributes.md
```diff
@@ -0,0 +1,33 @@
...
+
+All attributes are optional.
+
+| Attribute name       | Description            | Example value                   |
+|----------------------|------------------------|---------------------------------|
+| "http.host"          | Request URL host       | "example.com"                   |
+| "http.method"        | Request URL method     | "GET"                           |
+| "http.path"          | Request URL path       | "/users"                        |
+| "http.user_agent"    | Request user-agent     | "HTTPClient/1.2"                |
+| "http.status"        | Response status code   | 200                             |
```
bogdandrutu: I would name this status_code

trace/HTTPAttributes.md
```diff
@@ -0,0 +1,33 @@
...
+| "http.request_size"  | Request size in bytes  | 64328                           |
```
bogdandrutu: I think in our data model the sizes should be recorded via MessageEvents

adriancole:
Just to suggest another, which is a template which can be shared with metrics, usually associated with a symbolic route. Micrometer uses this in combination of status information to create a low cardinality grouping tag which is still identifiable. Others use this as an alternate to the span name (usually low cardinality span names default to the method). It could be early to add here, but in zipkin this has been discussed for quite a while and being implemented.

Here are some examples of http template values for the same endpoint, showcasing how they can vary in a framework specific way, but still accomplish low cardinality goals:

* vert.x web route "/items/:itemId"
* Play route "/items/$id<[0-9]+>"
* Spring WebMVC request mapping "/items/{itemId}"

https://github.com/openzipkin/zipkin/issues/1874

adriancole:
>
> Due to the security concerns, we should prefer having host and path instead of requiring URL to be recorded. I removed the URL attribute and also marked all attributes as optional. PTAL.
>
> ack. should someone need to re-assemble them later, it would be a bit
difficult, but anyway nothing prevents using custom tags in the mean time

ex. below from scala (where uri includes query params)

override def url(req: RequestHeader) =
  "http" + (if (req.secure) "s" else "") + "://" + req.host + req.uri

trace/HTTPAttributes.md
rakyll: You were reviewing an old version of the spec, this is renamed.

rakyll:
> Here are some examples of http template values for the same endpoint, showcasing how they can vary in a framework specific way, but still accomplish low cardinality goals:

I added an attribute to represent the matched route. We cannot enforce the a format due to the different matching languages used by the frameworks. It is also optional if the handler cannot access to the route.

trace/HTTPAttributes.md
```diff
@@ -0,0 +1,35 @@
...
+and server spans. For a server, request represents the incoming request.
+For a client, request represents the outgoing request.
+
+All attributes are optional.
+
+| Attribute name            | Description                 | Example value                   |
+|---------------------------|-----------------------------|---------------------------------|
+| "http.host"               | Request URL host            | "example.com"                   |
+| "http.method"             | Request URL method          | "GET"                           |
+| "http.path"               | Request URL path            | "/users"                        |
```
adriancole: maybe this could hit home by using a path value that corresponds to the route. Ex. "/users/25f4c31d"

rakyll:
@adriancole Is there a reference doc for Zipkin attributes I can link to? 

@yurishkuro, similarly I am interested if there is a reference for Jaeger. 

I want to update the spec with Zipkin and Jaeger specifics before merging this PR.

yurishkuro:
@rakyll Jaeger follows OpenTracing [semantic conventions](https://github.com/opentracing/specification/blob/master/semantic_conventions.md).

jcchavezs:
@rakyll you can get an insight on zipkin specifics from https://zipkin.io/pages/instrumenting.html but there are old concepts like `binary annotations` which should be translated into `tags` for example.

adriancole:
Actually while not the best formatting in the world, here are the zipkin
tags (formerly known as binary annotation) definitions.

https://zipkin.io/public/thrift/v1/zipkinCore.html#Const_HTTP_HOST

We did this a while back in coordination with the distributed-tracing
google group. This was in response to even earlier discussions and trying
to wrangle what was done regardless of api agenda:
https://groups.google.com/forum/#!searchin/distributed-tracing/http.host%7Csort:date/distributed-tracing/dnzo02hO8KY/54vdbgxABwAJ

Yuri mentions opentracing.. many of them were sourced from above. Uber was
I think zipkin still at the time, or maybe Jaeger was being written or
something... anyway, most of the http tags were defined a long time ago
(based on what was used at the time including stackdriver). Obviously some
things changed later (you can look at the version history, notably they
later added db tags)

Zipkin's thrift constants have more advice and other things for reasons
including we define a query interface, and indexing is important. The
thrift constant docs are routinely copied to code comments even if nobody
uses thrift anymore as a primary model.

The http.route tag is still being defined, I think you have the link to the
discussion, but here's the current comments for that.

/**
 * An expression such as "/items/:itemId", representing an application
endpoint, used to group
 * similar requests together. This can be absent on redirect,
file-not-found, or when the
 * framework doesn't include template expressions.
 *
 * <p>For example, the route "/products/{key}", would match
"/products/1" and "/products/2".
 * There is no format required for the encoding, as it is sometimes
application defined. The
 * important part is that the value namespace is low cardinality.
 *
 * <p>Conventionally associated with the key "http.route"
 *
 * <p>Eventhough the route is associated with the request, not the
response, this is present
 * on the response object. The reasons is that many server
implementations process the request
 * before they can identify the route route.
 */


Since OT mentions other things, it might be curious later:

sql - While not discussed in depth, we are likely to re-use sql tags from
xray as that's where the demand has been sourced from (we only have
sql.query at the moment)
https://docs.aws.amazon.com/xray/latest/devguide/xray-api-segmentdocuments.html#api-segmentdocuments-sql

remote and local endpoint - zipkin has a primary model for endpoint data,
that's visible in the openapi spec with docs
https://zipkin.io/zipkin-api/#/default/post_spans
https://zipkin.io/zipkin-api/zipkin2-api.yaml

PR #51, opened 2018-02-17
Initial Stats API overview.

bogdandrutu:
An overview of all three major components of the Stats API.

What is next:
* Implementation details/design.
* Performance requirements.
* Explain how can open-census support recording measurements that are not event based such as task_memory_usage, task_cpu_usage.

### DIFF
--- a/stats/DataAggregation.md
+++ b/stats/DataAggregation.md
@@ -1,0 +1,69 @@
+# Data Aggregation API Overview
+The stats library MUST allow users to define measurements how data are aggregated and broken down.
+The core data types used are:
+* `Aggregation`: defines the aggregation type (e.g. `Distribution`, `Sum`, `Mean`).
+* `View`: defines how individual measurements are broken down and aggregated.
+
+## Aggregation
+An Aggregation describes how data collected is aggregated.
+
+The library MUST provide support for multiple types of Aggregations:
+* `Count`: indicates that data collected and aggregated with this `Aggregation` type will be turned
+into a count value.
+* `Sum`: indicates that data collected and aggregated with this `Aggregation` will be summed up.
+* `Mean`: indicates that data collected and aggregated with this `Aggregation` will calculate the
+mean value.
+* `Distribution`: indicates that the desired `Aggregation` is a histogram distribution. A
+distribution `Aggregation` may contain a histogram of the values in the population. User MUST
+define the bucket boundaries for that histogram.
+
+## View
+A View describes how individual measurements are both broken down (by a set of tag keys, which
+form the basis for a Metric labels.) and aggregated (e.g. into Distributions). A single Measure
+can be used by multiple Views : this allows for different aggregations and uses of a single base
+type.
+
+A View is defined from the following:
+* `name`: a string by which the View will be referred to, e.g. "rpc_latency". Names MUST be unique
+within the library.
+* `description`: a free format descriptive string, e.g. "RPC counts for last minute and hour"
+* `Measure`: the Measure to which this view is applied. This field MAY be replaced with
+`MeasureDescription` if defined.
+* `columns`: an array of TagKeys. These values associated with tags of this name and type
+form the basis by which individual stats will be aggregated (one aggregation per unique tag value).
+If none are provided, then all data is recorded in a single aggregation.
+* `aggregation`: `Distribution`, `Count`, `Sum`, `Mean`.
+
+Implementations MUST define a View class, constructed from the parameters above. Views MAY have
+getters for retrieving all of the information used in View definition. Once created, View metadata
+is immutable.
+
+Example in Java
+```
+// Define list of tags used to break down view - just rpc method in this case
+ArrayList tagKeyListForMethodViews = new ArrayList();
+tagKeyListForMethodViews.add(rpcMethodKey);
+
+// Latencies from 100us to 10s.
+BucketBoundaries latencyDistributionBuckets =
+    new BucketBoundaries(Arrays.asList(0, 0.01, 0.1, 1.0, 10.0, 1000.0, 10000.0));
+
+// We want a Distribution as well.
+Distribution latencyDistribution = new Distribution.create(latencyDistributionBuckets);
+
+// Define the view with full stats
+View rpcLatencyByMethodView = new View(
+  "rpc_latency_by_method", // name
+  "Records full stats for RPC latency broken down by Method", // description
+  rpc_latency, // Measure
+  tagKeyListForMethodViews, // aggregation_tags
+  latencyDistribution);
+```
+
+References to Views in the system MAY be obtained from querying of registered Views. This
+functionality is required to decouple the definition of the data aggregation from the exporting
+of the aggregated data.
+
+Implementations MAY define a `ViewDescription` class which contains all the read-only fields from
+the `View` definition such as: `name`, `description`, `MeasureDescription` and `columns`
+and `Aggregation`.
--- a/stats/DataAggregation.md
+++ b/stats/DataAggregation.md
@@ -1,15 +1,15 @@
 # Data Aggregation API Overview
-The stats library allows users to define measurements how data are aggregated and broken down.
+The stats library allows users to define views to configure how data are aggregated and broken down.
 The core data types used are:
 * `Aggregation`: defines the aggregation type (e.g. `Distribution`, `Sum`, `Mean`).
-* `View`: defines how individual measurements are broken down by tags and aggregated.
+* `View`: defines how individual measurements are broken down by tags (similar to labels in some 
+other systems) and aggregated.
 
 ## Aggregation
 An Aggregation describes how data collected is aggregated.
 
 The library SHOULD provide support for multiple types of Aggregations:
-* `Count`: indicates that data collected and aggregated with this `Aggregation` type will be turned
-into a count value.
+* `Count`: counts the number of measurements recorded.
 * `Sum`: indicates that data collected and aggregated with this `Aggregation` will be summed up.
 * `Mean`: indicates that data collected and aggregated with this `Aggregation` will calculate the
 mean value.
@@ -18,18 +18,17 @@ distribution `Aggregation` may contain a histogram of the values in the populati
 define the bucket boundaries for that histogram.
 
 ## View
-A View describes how individual measurements are both broken down (by a set of tag keys, which
-form the basis for a Metric labels.) and aggregated (e.g. into Distributions). A single Measure
-can be used by multiple Views.
+A View describes how individual measurements are both broken down (by a set of tag keys) and 
+aggregated (e.g. into Distributions). A single `Measure` can be used by multiple Views.
 
 A View is defined from the following:
 * `name`: a string by which the View will be referred to, e.g. "rpc_latency". Names MUST be unique
 within the library.
 * `description`: a free format descriptive string, e.g. "RPC counts for last minute and hour"
 * `Measure`: the Measure to which this view is applied.
-* `columns`: an array of tag keys. These values associated with tags of this name and type
-form the basis by which individual stats will be aggregated (one aggregation per unique tag value).
-If none are provided, then all data is recorded in a single aggregation.
+* `columns`: an array of tag keys. These values associated with tags of this name form the basis 
+by which individual stats will be aggregated (one aggregation per unique tag value). If none are 
+provided, then all data is recorded in a single aggregation.
 * `aggregation`: `Distribution`, `Count`, `Sum`, `Mean`.
 
 Implementations SHOULD define a View data type, constructed from the parameters above. Views MAY 
@@ -37,20 +36,21 @@ have getters for retrieving all of the information used in View definition. Once
 metadata is immutable.
 
 Example in Java
-```
+```java
 // Define a list of tags to break down view by RPC method.
-ArrayList tagKeyListForMethodViews = new ArrayList();
+private static final ArrayList tagKeyListForMethodViews = new ArrayList();
 tagKeyListForMethodViews.add(RPC_METHOD_KEY);
 
 // Create latency buckets from 100us to 10s.
-BucketBoundaries LATENCY_DISTRIBUTION_BUCKETS =
+private static final BucketBoundaries LATENCY_DISTRIBUTION_BUCKETS =
     new BucketBoundaries(Arrays.asList(0, 0.01, 0.1, 1.0, 10.0, 1000.0, 10000.0));
 
 // Create a distribution with the latency buckets.
-Distribution LATENCY_DISTRIBUTION = new Distribution.create(LATENCY_DISTRIBUTION_BUCKETS);
+private static final Distribution LATENCY_DISTRIBUTION =
+    Distribution.create(LATENCY_DISTRIBUTION_BUCKETS);
 
 // Define a view to collect RPC latency stats by RPC method.
-View rpcLatencyByMethodView = new View(
+private static final View RPC_LATENCY_BY_METHOD_VIEW = new View(
   "rpc_latency_by_method", // name
   "Distribution of RPC latency broken down by Method", // description
   RPC_LATENCY, // Measure
@@ -62,6 +62,6 @@ References to Views MAY be obtained from querying of registered Views at runtime
 functionality enables decoupling the definition of the data aggregation from the exporting
 of the aggregated data.
 
-Implementations MAY define a `ViewDescription` data type which contains all the read-only fields 
-from the `View` definition such as: `name`, `description`, `MeasureDescription` and `columns`
-and `Aggregation`.
+For languages that do not allow private properties/metadata and if they are needed implementations 
+MAY define a `ViewDescription` data type which contains all the read-only fields from the `View` 
+definition such as: `name`, `description`, `MeasureDescription` and `columns` and `Aggregation`.
--- a/stats/DataAggregation.md
+++ b/stats/DataAggregation.md
@@ -13,6 +13,10 @@ The library SHOULD provide support for multiple types of Aggregations:
 * `Sum`: indicates that data collected and aggregated with this `Aggregation` will be summed up.
 * `Mean`: indicates that data collected and aggregated with this `Aggregation` will calculate the
 mean value.
+* `Max`: indicates that data collected and aggregated with this `Aggregation` will calculate the 
+maximum value recorded.
+* `LastValue`: indicates that data collected and aggregated with this `Aggregation` will 
+represent the last recorded value. This is useful to support Gauges.
 * `Distribution`: indicates that the desired `Aggregation` is a histogram distribution. A
 distribution `Aggregation` may contain a histogram of the values in the population. User should
 define the bucket boundaries for that histogram.
--- a/stats/DataAggregation.md
+++ b/stats/DataAggregation.md
@@ -14,7 +14,7 @@ into a count value.
 * `Mean`: indicates that data collected and aggregated with this `Aggregation` will calculate the
 mean value.
 * `Distribution`: indicates that the desired `Aggregation` is a histogram distribution. A
-distribution `Aggregation` may contain a histogram of the values in the population. User can
+distribution `Aggregation` may contain a histogram of the values in the population. User should
 define the bucket boundaries for that histogram.
 
 ## View
@@ -40,22 +40,22 @@ Example in Java
 ```
 // Define a list of tags to break down view by RPC method.
 ArrayList tagKeyListForMethodViews = new ArrayList();
-tagKeyListForMethodViews.add(rpcMethodKey);
+tagKeyListForMethodViews.add(RPC_METHOD_KEY);
 
 // Create latency buckets from 100us to 10s.
-BucketBoundaries latencyDistributionBuckets =
+BucketBoundaries LATENCY_DISTRIBUTION_BUCKETS =
     new BucketBoundaries(Arrays.asList(0, 0.01, 0.1, 1.0, 10.0, 1000.0, 10000.0));
 
 // Create a distribution with the latency buckets.
-Distribution latencyDistribution = new Distribution.create(latencyDistributionBuckets);
+Distribution LATENCY_DISTRIBUTION = new Distribution.create(LATENCY_DISTRIBUTION_BUCKETS);
 
 // Define a view to collect RPC latency stats by RPC method.
 View rpcLatencyByMethodView = new View(
   "rpc_latency_by_method", // name
   "Distribution of RPC latency broken down by Method", // description
-  rpcLatency, // Measure
+  RPC_LATENCY, // Measure
   tagKeyListForMethodViews, // aggregation_tags
-  latencyDistribution);
+  LATENCY_DISTRIBUTION);
 ```
 
 References to Views MAY be obtained from querying of registered Views at runtime. This
--- a/stats/DataAggregation.md
+++ b/stats/DataAggregation.md
@@ -11,8 +11,6 @@ An Aggregation describes how data collected is aggregated.
 The library SHOULD provide support for multiple types of Aggregations:
 * `Count`: counts the number of measurements recorded.
 * `Sum`: indicates that data collected and aggregated with this `Aggregation` will be summed up.
-* `Mean`: indicates that data collected and aggregated with this `Aggregation` will calculate the
-mean value.
 * `Max`: indicates that data collected and aggregated with this `Aggregation` will calculate the 
 maximum value recorded.
 * `LastValue`: indicates that data collected and aggregated with this `Aggregation` will 
--- a/stats/DataAggregation.md
+++ b/stats/DataAggregation.md
@@ -1,69 +1,67 @@
 # Data Aggregation API Overview
-The stats library MUST allow users to define measurements how data are aggregated and broken down.
+The stats library allows users to define measurements how data are aggregated and broken down.
 The core data types used are:
 * `Aggregation`: defines the aggregation type (e.g. `Distribution`, `Sum`, `Mean`).
-* `View`: defines how individual measurements are broken down and aggregated.
+* `View`: defines how individual measurements are broken down by tags and aggregated.
 
 ## Aggregation
 An Aggregation describes how data collected is aggregated.
 
-The library MUST provide support for multiple types of Aggregations:
+The library SHOULD provide support for multiple types of Aggregations:
 * `Count`: indicates that data collected and aggregated with this `Aggregation` type will be turned
 into a count value.
 * `Sum`: indicates that data collected and aggregated with this `Aggregation` will be summed up.
 * `Mean`: indicates that data collected and aggregated with this `Aggregation` will calculate the
 mean value.
 * `Distribution`: indicates that the desired `Aggregation` is a histogram distribution. A
-distribution `Aggregation` may contain a histogram of the values in the population. User MUST
+distribution `Aggregation` may contain a histogram of the values in the population. User can
 define the bucket boundaries for that histogram.
 
 ## View
 A View describes how individual measurements are both broken down (by a set of tag keys, which
 form the basis for a Metric labels.) and aggregated (e.g. into Distributions). A single Measure
-can be used by multiple Views : this allows for different aggregations and uses of a single base
-type.
+can be used by multiple Views.
 
 A View is defined from the following:
 * `name`: a string by which the View will be referred to, e.g. "rpc_latency". Names MUST be unique
 within the library.
 * `description`: a free format descriptive string, e.g. "RPC counts for last minute and hour"
-* `Measure`: the Measure to which this view is applied. This field MAY be replaced with
-`MeasureDescription` if defined.
-* `columns`: an array of TagKeys. These values associated with tags of this name and type
+* `Measure`: the Measure to which this view is applied.
+* `columns`: an array of tag keys. These values associated with tags of this name and type
 form the basis by which individual stats will be aggregated (one aggregation per unique tag value).
 If none are provided, then all data is recorded in a single aggregation.
 * `aggregation`: `Distribution`, `Count`, `Sum`, `Mean`.
 
-Implementations MUST define a View class, constructed from the parameters above. Views MAY have
-getters for retrieving all of the information used in View definition. Once created, View metadata
-is immutable.
+Implementations SHOULD define a View data type, constructed from the parameters above. Views MAY 
+have getters for retrieving all of the information used in View definition. Once created, View 
+metadata is immutable.
 
 Example in Java
 ```
-// Define list of tags used to break down view - just rpc method in this case
+// Define a list of tags to break down view by RPC method.
 ArrayList tagKeyListForMethodViews = new ArrayList();
 tagKeyListForMethodViews.add(rpcMethodKey);
 
-// Latencies from 100us to 10s.
+// Create latency buckets from 100us to 10s.
 BucketBoundaries latencyDistributionBuckets =
     new BucketBoundaries(Arrays.asList(0, 0.01, 0.1, 1.0, 10.0, 1000.0, 10000.0));
 
-// We want a Distribution as well.
+// Create a distribution with the latency buckets.
 Distribution latencyDistribution = new Distribution.create(latencyDistributionBuckets);
 
-// Define the view with full stats
+// Define a view to collect RPC latency stats by RPC method.
 View rpcLatencyByMethodView = new View(
   "rpc_latency_by_method", // name
-  "Records full stats for RPC latency broken down by Method", // description
-  rpc_latency, // Measure
+  "Distribution of RPC latency broken down by Method", // description
+  rpcLatency, // Measure
   tagKeyListForMethodViews, // aggregation_tags
   latencyDistribution);
 ```
 
-References to Views in the system MAY be obtained from querying of registered Views. This
-functionality is required to decouple the definition of the data aggregation from the exporting
+References to Views MAY be obtained from querying of registered Views at runtime. This
+functionality enables decoupling the definition of the data aggregation from the exporting
 of the aggregated data.
 
-Implementations MAY define a `ViewDescription` class which contains all the read-only fields from
-the `View` definition such as: `name`, `description`, `MeasureDescription` and `columns`
+Implementations MAY define a `ViewDescription` data type which contains all the read-only fields 
+from the `View` definition such as: `name`, `description`, `MeasureDescription` and `columns`
 and `Aggregation`.
--- a/stats/DataAggregation.md
+++ b/stats/DataAggregation.md
@@ -24,7 +24,7 @@ aggregated (e.g. into Distributions). A single `Measure` can be used by multiple
 A View is defined from the following:
 * `name`: a string by which the View will be referred to, e.g. "rpc_latency". Names MUST be unique
 within the library.
-* `description`: a free format descriptive string, e.g. "RPC counts for last minute and hour"
+* `description`: a free format descriptive string, e.g. "RPC latency distribution".
 * `Measure`: the Measure to which this view is applied.
 * `columns`: an array of tag keys. These values associated with tags of this name form the basis 
 by which individual stats will be aggregated (one aggregation per unique tag value). If none are 
@@ -51,7 +51,7 @@ private static final Distribution LATENCY_DISTRIBUTION =
 
 // Define a view to collect RPC latency stats by RPC method.
 private static final View RPC_LATENCY_BY_METHOD_VIEW = new View(
-  "rpc_latency_by_method", // name
+  View.Name.create("rpc_latency_by_method"),
   "Distribution of RPC latency broken down by Method", // description
   RPC_LATENCY, // Measure
   tagKeyListForMethodViews, // aggregation_tags
--- a/stats/Export.md
+++ b/stats/Export.md
@@ -14,6 +14,8 @@ The library SHOULD provide support for multiple types of Aggregations:
 * `SumDataDouble` and `SumDataInt64`: data generated for a `Sum` aggregation based on the `Measure`
 type.
 * `MeanData`: data generated for a `Mean` aggregation.
+* `MaxData`: data generated for a `Max` aggregation.
+* `LastValueData`: data generated for a `LastValue` aggregation.
 * `DistributionData`: data generated for a `Distribution` aggregation.
 
 ### ViewData
--- a/stats/Export.md
+++ b/stats/Export.md
@@ -1,7 +1,7 @@
 # Export API Overview
 The stats library aggregates measurements into Views (which are essentially a collection of
-metrics, each with a different set of labels). All output objects are immutable and read-only.
-The core data types used are:
+metrics, each with a different set of labels). All output objects are immutable. The core data 
+types used are:
 * `AggregationData`: contains data from an `Aggregation`.
 * `ViewData`: encapsulates `View` data aggregation.
 
@@ -11,7 +11,7 @@ AggregationData for each Aggregation types defined.
 
 The library SHOULD provide support for multiple types of Aggregations:
 * `CountData`: data generated for a `Count` aggregation.
-* `SumDataDouble` and `SumDataInt`: data generated for a `Sum` aggregation based on the `Measure`
+* `SumDataDouble` and `SumDataInt64`: data generated for a `Sum` aggregation based on the `Measure`
 type.
 * `MeanData`: data generated for a `Mean` aggregation.
 * `DistributionData`: data generated for a `Distribution` aggregation.
--- a/stats/Export.md
+++ b/stats/Export.md
@@ -18,17 +18,11 @@ type.
 
 ### ViewData
 A ViewData is defined from the following:
-* `name`: the name of the exported `View`.
-* `columns`: the `columns` of the exported `View`.
-* `Measure`: the `Measure` of the exported `View`. This field MAY be replaced with
-`MeasureDescription` if defined.
-* `aggregation`: the `aggregation` of the exported `View`
+* `View`: the exported `View`.
 * `rows`: a map of repeated tag values, and, dependent on Aggregation type in the `View` an
 `AggregationData` object. These contain the respective tag values corresponding to the `columns`
 parameter from the `View` and the aggregated data associated with those `columns`.
-* `start_time`: a timestamp, describing the start time at which the underlying View started
-collection of these data.
-* `end_time`: a timestamp, describing the end time at which the underlying View ended collection
-of these data.
+* `start_time`: a timestamp, describing the start time of the current stats snapshot.
+* `end_time`: a timestamp, describing the end time of the current stats snapshot.
 
 The library SHOULD provide a means of retrieving the ViewData for any registered view in the system.
--- a/stats/Export.md
+++ b/stats/Export.md
@@ -1,0 +1,34 @@
+# Export API Overview
+The stats library aggregates measurements into Views (which are essentially a collection of
+metrics, each with a different set of labels). All output objects are immutable and read-only.
+The core data types used are:
+* `AggregationData`: contains data from an `Aggregation`.
+* `ViewData`: encapsulates `View` data aggregation.
+
+### AggregationData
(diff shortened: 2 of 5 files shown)


stats/DataAggregation.md
```diff
@@ -0,0 +1,69 @@
...
+  rpc_latency, // Measure
+  tagKeyListForMethodViews, // aggregation_tags
+  latencyDistribution);
+```
+
+References to Views in the system MAY be obtained from querying of registered Views. This
+functionality is required to decouple the definition of the data aggregation from the exporting
+of the aggregated data.
+
+Implementations MAY define a `ViewDescription` class which contains all the read-only fields from
```
ramonza: again, a class is not necessarily best here for many languages

stats/README.md
```diff
@@ -0,0 +1,14 @@
+# OpenCensus Library Trace API
```
rakyll: Stats API

stats/README.md
```diff
@@ -0,0 +1,14 @@
...
+This document serves to document the "look and feel" of the open source stats API's. It describes
+the key types (classes, structures, etc.) and functions (methods, etc.).
```
rakyll: You can say key types and the behavior.

stats/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+OpenCensus allows users to create typed measures, record measurements, define data aggregation, and
+export aggregated data.
+
+## Main APIs
+* [Record API](Record.md): used by the application and third-party libraries (e.g gRPC) owners to
+record metrics.
+* [Data Aggregation API](DataAggregation.md): used by the application owners to define data
```
rakyll: I'd s/to define/to define and enable

stats/Record.md
```diff
@@ -0,0 +1,73 @@
...
+## Measure
+A `Measure` describes the type of the individual values/measurements recorded by an application. It
+includes information such as the type of measurement, the units of measurement and descriptive names
+for the data. This provides the fundamental type used for recording data.
+
+A Measure describes a value with the following metadata:
+* `name`: a string by which the measure will be referred to, e.g. "rpc_server_latency", or
+"vm_cpu_cycles". Names MUST be unique within the library. It is recommended to use names compatible
+with the intended end usage, e.g, use host/path pattern.
+* `description`: a free format descriptive string, e.g. "RPC latency in seconds", "Virtual cycles
```
rakyll: : a string describes the measure

stats/Record.md
```diff
@@ -0,0 +1,73 @@
...
+executed on VM".
+* `unit`: a string describing the unit's used for this Measure. Follows the format described by
```
rakyll: a string describing the unit used for the measure

stats/Record.md
```diff
@@ -0,0 +1,73 @@
...
+Implementations MAY define a `MeasurementMap` which describes a set of data points to be collected
+for a set of Measures. Adding this functionality may improve the record usage API.
+
+## Record usage
+
+Users MUST record Measurements with the current context (implicit or explicit). Tags from the
+current context are recorded with the Measurements if they are any.
+
+Implementations MUST provide a means of recording Measurements. This functionality SHOULD be
+provided through one of the following options:
```
rakyll: We need to encourage recording multiple data points at a single time to reduce the deserialization cost.

Shouldn't we require the first option as a MUST and the second one optionally?

stats/DataAggregation.md
```diff
@@ -0,0 +1,69 @@
...
+form the basis for a Metric labels.) and aggregated (e.g. into Distributions). A single Measure
+can be used by multiple Views : this allows for different aggregations and uses of a single base
+type.
+
+A View is defined from the following:
+* `name`: a string by which the View will be referred to, e.g. "rpc_latency". Names MUST be unique
+within the library.
+* `description`: a free format descriptive string, e.g. "RPC counts for last minute and hour"
+* `Measure`: the Measure to which this view is applied. This field MAY be replaced with
+`MeasureDescription` if defined.
```
rakyll: I'd not mention MeasureDescription, it is more of an implementation detail than the general rule.

stats/DataAggregation.md
```diff
@@ -0,0 +1,69 @@
...
+* `columns`: an array of TagKeys. These values associated with tags of this name and type
```
rakyll: array of tag keys. TagKey sounds like a Java class name.

stats/DataAggregation.md
```diff
@@ -0,0 +1,69 @@
...
+If none are provided, then all data is recorded in a single aggregation.
+* `aggregation`: `Distribution`, `Count`, `Sum`, `Mean`.
+
+Implementations MUST define a View class, constructed from the parameters above. Views MAY have
+getters for retrieving all of the information used in View definition. Once created, View metadata
+is immutable.
+
+Example in Java
+```
+// Define list of tags used to break down view - just rpc method in this case
```
rakyll: // Define a list of tags to break down view by RPC method.

stats/DataAggregation.md
```diff
@@ -0,0 +1,69 @@
...
+ArrayList tagKeyListForMethodViews = new ArrayList();
+tagKeyListForMethodViews.add(rpcMethodKey);
+
+// Latencies from 100us to 10s.
```
rakyll: Create latency buckets...

stats/DataAggregation.md
```diff
@@ -0,0 +1,69 @@
...
+BucketBoundaries latencyDistributionBuckets =
+    new BucketBoundaries(Arrays.asList(0, 0.01, 0.1, 1.0, 10.0, 1000.0, 10000.0));
+
+// We want a Distribution as well.
+Distribution latencyDistribution = new Distribution.create(latencyDistributionBuckets);
```
rakyll: // Create a distribution with the latency buckets.

stats/DataAggregation.md
```diff
@@ -0,0 +1,69 @@
...
+
+// Define the view with full stats
```
rakyll: // Define a view to collect RPC latency stats by RPC method.

stats/DataAggregation.md
```diff
@@ -0,0 +1,69 @@
...
+View rpcLatencyByMethodView = new View(
+  "rpc_latency_by_method", // name
+  "Records full stats for RPC latency broken down by Method", // description
```
rakyll: I'd avoid the use of "full". It doesn't add and confuses the reader. "Is there a way to read stats partially?"

rakyll: Remove the mention of "Records". It is open to confusion given recording means something specific in OpenCensus.

stats/DataAggregation.md
```diff
@@ -0,0 +1,69 @@
...
+  rpc_latency, // Measure
```
rakyll: rpcLatency

stats/DataAggregation.md
```diff
@@ -0,0 +1,69 @@
...
+  tagKeyListForMethodViews, // aggregation_tags
+  latencyDistribution);
+```
+
+References to Views in the system MAY be obtained from querying of registered Views. This
```
rakyll: What's the system here?

You can either say, "References to Views MAY be obtained... in runtime."

stats/DataAggregation.md
```diff
@@ -0,0 +1,69 @@
+# Data Aggregation API Overview
+The stats library MUST allow users to define measurements how data are aggregated and broken down.
+The core data types used are:
+* `Aggregation`: defines the aggregation type (e.g. `Distribution`, `Sum`, `Mean`).
+* `View`: defines how individual measurements are broken down and aggregated.
```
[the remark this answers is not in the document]
bogdandrutu: Done.

stats/DataAggregation.md
```diff
@@ -0,0 +1,69 @@
...
+mean value.
+* `Distribution`: indicates that the desired `Aggregation` is a histogram distribution. A
+distribution `Aggregation` may contain a histogram of the values in the population. User MUST
+define the bucket boundaries for that histogram.
+
+## View
+A View describes how individual measurements are both broken down (by a set of tag keys, which
+form the basis for a Metric labels.) and aggregated (e.g. into Distributions). A single Measure
+can be used by multiple Views : this allows for different aggregations and uses of a single base
+type.
```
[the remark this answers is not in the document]
bogdandrutu: Done.

stats/DataAggregation.md
bogdandrutu: done.

stats/DataAggregation.md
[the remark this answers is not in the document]
bogdandrutu: A tag is a key->value and here users have to define only the keys and maybe mutations (TagSelector). So to answer your first question tags is wrong here. I'd add tag selector later in a different PR after we document the current state.

stats/DataAggregation.md
```diff
@@ -0,0 +1,69 @@
...
+within the library.
+* `description`: a free format descriptive string, e.g. "RPC counts for last minute and hour"
+* `Measure`: the Measure to which this view is applied. This field MAY be replaced with
+`MeasureDescription` if defined.
+* `columns`: an array of TagKeys. These values associated with tags of this name and type
+form the basis by which individual stats will be aggregated (one aggregation per unique tag value).
+If none are provided, then all data is recorded in a single aggregation.
+* `aggregation`: `Distribution`, `Count`, `Sum`, `Mean`.
+
+Implementations MUST define a View class, constructed from the parameters above. Views MAY have
```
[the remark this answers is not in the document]
bogdandrutu: Changed class with "data type"

stats/DataAggregation.md
bogdandrutu: done.

stats/DataAggregation.md
bogdandrutu: done.

stats/DataAggregation.md
bogdandrutu: done.

stats/DataAggregation.md
bogdandrutu: Done.

stats/DataAggregation.md
bogdandrutu: done.

stats/DataAggregation.md
bogdandrutu: agreed.

bogdandrutu: done.

stats/DataAggregation.md
bogdandrutu: done.

stats/DataAggregation.md
bogdandrutu: done.

stats/DataAggregation.md
bogdandrutu: Done. "data type"

stats/README.md
bogdandrutu: done.

stats/Record.md
bogdandrutu: done.

stats/README.md
bogdandrutu: Done.

stats/README.md
bogdandrutu: done.

stats/Record.md
bogdandrutu: done.

stats/Record.md
bogdandrutu: The second is still batch. MeasurementMap is essentially a collection of Measurements. See the example in Java.

stats/DataAggregation.md
```diff
@@ -0,0 +1,69 @@
...
+A View describes how individual measurements are both broken down (by a set of tag keys, which
+form the basis for a Metric labels.) and aggregated (e.g. into Distributions). A single Measure
+can be used by multiple Views : this allows for different aggregations and uses of a single base
+type.
+
+A View is defined from the following:
+* `name`: a string by which the View will be referred to, e.g. "rpc_latency". Names MUST be unique
+within the library.
+* `description`: a free format descriptive string, e.g. "RPC counts for last minute and hour"
+* `Measure`: the Measure to which this view is applied. This field MAY be replaced with
```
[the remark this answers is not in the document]
bogdandrutu: We can do that iff we require that all implementations enforce uniqueness of the Measure name.

stats/DataAggregation.md
```diff
@@ -0,0 +1,69 @@
...
+* `Sum`: indicates that data collected and aggregated with this `Aggregation` will be summed up.
+* `Mean`: indicates that data collected and aggregated with this `Aggregation` will calculate the
+mean value.
+* `Distribution`: indicates that the desired `Aggregation` is a histogram distribution. A
+distribution `Aggregation` may contain a histogram of the values in the population. User MUST
+define the bucket boundaries for that histogram.
+
+## View
+A View describes how individual measurements are both broken down (by a set of tag keys, which
+form the basis for a Metric labels.) and aggregated (e.g. into Distributions). A single Measure
```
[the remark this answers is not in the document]
bogdandrutu: In the metric systems usually the common word is label. Here I tried to show people the correlation between OpenCensus tags and Metric label.

bogdandrutu:
@ramonza I tried to relax the requirements and try to make some of the comments look more like suggestions than directions (still probably a bit more direct than you want). Do you think we are moving to the right direction?

Some changes:
* Replace few MUST with SHOULD.
* Replace "class" with "data type".

stats/DataAggregation.md
```diff
@@ -0,0 +1,67 @@
...
+## Aggregation
+An Aggregation describes how data collected is aggregated.
+
+The library SHOULD provide support for multiple types of Aggregations:
+* `Count`: indicates that data collected and aggregated with this `Aggregation` type will be turned
+into a count value.
+* `Sum`: indicates that data collected and aggregated with this `Aggregation` will be summed up.
+* `Mean`: indicates that data collected and aggregated with this `Aggregation` will calculate the
+mean value.
+* `Distribution`: indicates that the desired `Aggregation` is a histogram distribution. A
```
isturdy: Stackdriver Distribution includes mean, sum of squared deviation, and range; C++ and Java include those in their Distribution type. To reflect existing implementations I would make that "Distribution", and possibly include a histogram-only "Histogram" type.

stats/Record.md
```diff
@@ -0,0 +1,73 @@
+# Record API Overview
+The stats library allows users to record metrics for their applications or third-party libraries.
+ The core data types used are:
+* `Measure`: describes the type of the individual values recorded by an application.
+* `Measurement`: describes a data point to be collected for a `Measure`.
+
+## Measure
+A `Measure` describes the type of the individual values/measurements recorded by an application. It
+includes information such as the type of measurement, the units of measurement and descriptive names
```
songy23: nit: Measure also has a brief description.

stats/DataAggregation.md
```diff
@@ -0,0 +1,67 @@
+# Data Aggregation API Overview
+The stats library allows users to define measurements how data are aggregated and broken down.
+The core data types used are:
```
isturdy: I still think we should leave Aggregation Window as a concept; even if we should leave it out of the public API for simplicity, z-page and LB support make it likely that multiple implementations will want to keep the concept internally, and if so it is still useful to standardize terminology.

stats/Record.md
```diff
@@ -0,0 +1,73 @@
...
+* `name`: a string by which the measure will be referred to, e.g. "rpc_server_latency", or
+"vm_cpu_cycles". Names MUST be unique within the library. It is recommended to use names 
+compatible with the intended end usage, e.g, use host/path pattern.
+* `description`: a string describes the measure, e.g. "RPC latency in seconds", "Virtual cycles
+executed on VM".
+* `unit`: a string describing the unit used for the 'Measure. Follows the format described by
+[Unified Code for Units of Measure](http://unitsofmeasure.org/ucum.html).
+* `type`: the only supported types are `int64` and `double`.
+
+Implementations SHOULD define a `Measure` data taype, constructed from the parameters above.
```
songy23: s/taype/type

stats/Record.md
```diff
@@ -0,0 +1,73 @@
...
+[Unified Code for Units of Measure](http://unitsofmeasure.org/ucum.html).
+* `type`: the only supported types are `int64` and `double`.
+
+Implementations SHOULD define a `Measure` data taype, constructed from the parameters above. 
+Measure MAY have getters for retrieving all of the information used in `Measure` definition. 
+Once created, Measure metadata is immutable.
+
+Example in Java
+```
+private static final MeasureDouble rpcLatency =
```
songy23: nit: s/rpcLatency/RPC_LATENCY, since it's an immutable constant.

stats/Record.md
```diff
@@ -0,0 +1,73 @@
...
+    MeasureDouble.create("grpc.io/latency", "latency", "ms");
+```
+
+References to Measures in the system MAY be obtained from querying of registered Measures. This
+functionality is required to decouple the recording of the data from the exporting of the data.
+
+Example in Java
+```
+private static final MeasureDouble rpcLatency = Stats.findMeasureDoubleByName("grpc.io/latency");
```
songy23: Java doesn't have such an API, maybe using Go as example instead?

stats/Record.md
```diff
@@ -0,0 +1,73 @@
...
+A `Measurement` describes a value with the following metadata:
+* `Measure`: the `Measure` to which this `value` is applied.
+* `value`: recorded value, MUST have the appropriate type to match the `Measure` definition.
+
+Implementations MAY define a `MeasurementMap` which describes a set of data points to be collected
+for a set of Measures. Adding this functionality may improve the record usage API.
+
+## Record usage
+
+Users should record Measurements with the current context (implicit or explicit). Tags from the
```
songy23: I think users can record measurements against any context (doesn't necessarily to be current context): https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/stats/MeasureMap.java#L64.

stats/Record.md
```diff
@@ -0,0 +1,73 @@
...
+
+Implementations SHOULD provide a means of recording Measurements. This functionality MAY be
+provided through one of the following options:
+* As a method in a class/package (recommended name Stats/stats), taking a list of Measurement as
+argument. e.g. `record(List<Measurement>)` or `record(...Measurement)`.
+* As a `record` method of the appropriate data type. e.g. `MeasurementMap.record()`.
+
+Example in Java
+```
+private static final MeasureDouble rpcLatency =
```
songy23: Same, s/rpcLatency/RPC_LATENCY.

stats/Record.md
```diff
@@ -0,0 +1,73 @@
...
+    MeasureDouble.create("grpc.io/latency", "latency", "ms")
+private static final MeasureLong rpcBytesSent =
```
songy23: Same here.

stats/Export.md
```diff
@@ -0,0 +1,34 @@
...
+* `AggregationData`: contains data from an `Aggregation`.
+* `ViewData`: encapsulates `View` data aggregation.
+
+### AggregationData
+An AggregationData describes the result of data aggregation. The library SHOULD define an
+AggregationData for each Aggregation types defined.
+
+The library SHOULD provide support for multiple types of Aggregations:
+* `CountData`: data generated for a `Count` aggregation.
+* `SumDataDouble` and `SumDataInt`: data generated for a `Sum` aggregation based on the `Measure`
```
songy23: s/SumDataInt/SumDataInt64 or SumDataLong?

stats/DataAggregation.md
```diff
@@ -0,0 +1,67 @@
...
+* `Aggregation`: defines the aggregation type (e.g. `Distribution`, `Sum`, `Mean`).
+* `View`: defines how individual measurements are broken down by tags and aggregated.
+
+## Aggregation
+An Aggregation describes how data collected is aggregated.
+
+The library SHOULD provide support for multiple types of Aggregations:
```
isturdy: This requirement is technically satisfied by an implementation with any two Aggregations--perhaps we want to specify required and optional Aggregations? My inclination would be to make only a full-featured Distribution required, as it subsumes count, mean, range, and histogram, and sum unless full precision is required.

stats/Export.md
```diff
@@ -0,0 +1,34 @@
...
+* `columns`: the `columns` of the exported `View`.
+* `Measure`: the `Measure` of the exported `View`. This field MAY be replaced with
+`MeasureDescription` if defined.
+* `aggregation`: the `aggregation` of the exported `View`
+* `rows`: a map of repeated tag values, and, dependent on Aggregation type in the `View` an
+`AggregationData` object. These contain the respective tag values corresponding to the `columns`
+parameter from the `View` and the aggregated data associated with those `columns`.
+* `start_time`: a timestamp, describing the start time at which the underlying View started
+collection of these data.
+* `end_time`: a timestamp, describing the end time at which the underlying View ended collection
```
songy23: I think `end time` should be the timestamp when users retrieve data for the corresponding View, or say, the end time of current stats snapshot. Meanwhile stats collection is still ongoing.

stats/Export.md
```diff
@@ -0,0 +1,34 @@
+# Export API Overview
+The stats library aggregates measurements into Views (which are essentially a collection of
+metrics, each with a different set of labels). All output objects are immutable and read-only.
```
isturdy: Is not "immutable and read-only" redundant?

stats/DataAggregation.md
```diff
@@ -0,0 +1,67 @@
...
+* `Count`: indicates that data collected and aggregated with this `Aggregation` type will be turned
+into a count value.
+* `Sum`: indicates that data collected and aggregated with this `Aggregation` will be summed up.
+* `Mean`: indicates that data collected and aggregated with this `Aggregation` will calculate the
+mean value.
+* `Distribution`: indicates that the desired `Aggregation` is a histogram distribution. A
+distribution `Aggregation` may contain a histogram of the values in the population. User can
```
songy23: nit: User can -> User should

stats/DataAggregation.md
```diff
@@ -0,0 +1,67 @@
...
+define the bucket boundaries for that histogram.
+
+## View
+A View describes how individual measurements are both broken down (by a set of tag keys, which
+form the basis for a Metric labels.) and aggregated (e.g. into Distributions). A single Measure
+can be used by multiple Views.
+
+A View is defined from the following:
+* `name`: a string by which the View will be referred to, e.g. "rpc_latency". Names MUST be unique
```
songy23: Should we mention the constraint on View names? Currently it should be a ASCII string with a length <= 255 characters.

stats/Record.md
```diff
@@ -0,0 +1,73 @@
...
+* `Measure`: describes the type of the individual values recorded by an application.
+* `Measurement`: describes a data point to be collected for a `Measure`.
+
+## Measure
+A `Measure` describes the type of the individual values/measurements recorded by an application. It
+includes information such as the type of measurement, the units of measurement and descriptive names
+for the data. This provides the fundamental type used for recording data.
+
+A Measure describes a value with the following metadata:
+* `name`: a string by which the measure will be referred to, e.g. "rpc_server_latency", or
```
songy23: Same as View names. Measure names should be ASCII string with no more than 255 characters.

stats/Export.md
```diff
@@ -0,0 +1,34 @@
...
+The core data types used are:
+* `AggregationData`: contains data from an `Aggregation`.
+* `ViewData`: encapsulates `View` data aggregation.
+
+### AggregationData
+An AggregationData describes the result of data aggregation. The library SHOULD define an
+AggregationData for each Aggregation types defined.
```
isturdy: This does not make sense for all languages--I think C++ is cleaner with no AggregationData (ViewData contains one of several maps, each with a primitive mapped type), and even if it did adopt AggregationData it would be cleaner as a union type with different accessors rather than an interface and an implementation per Aggregation.

As Ramon noted, if we are going to make this a formal API specification, we need to be careful that we are only specifying things at a level of abstraction that translates well across languages--this section on AggregationData, in particular, seems to have Java-esque languages in mind.

stats/Export.md
isturdy: Probably Int64; Long is more language-specific (and in C++ is often only 32 bits).

stats/Record.md
isturdy: For that matter, there is no real concept of "current explicit context". I would favor "Users record Measurements against a context, either an explicit context or the implicit current context."

stats/Record.md
bogdandrutu: It is mentioned in the metadata section. Here is just an overview.

stats/Record.md
bogdandrutu: done.

stats/Export.md
bogdandrutu: done.

stats/DataAggregation.md
bogdandrutu: We will add a different doc explaining implementation high-level design which will address the window issue.

stats/Export.md
bogdandrutu: Done.

stats/DataAggregation.md
bogdandrutu: That should get into the DistributionData. The data that we produce is not part of this API.

stats/DataAggregation.md
bogdandrutu: done.

stats/DataAggregation.md
bogdandrutu: We can add later a section with restrictions for all types.

stats/Record.md
bogdandrutu: We can add later a section with restrictions for all types.

stats/Record.md
bogdandrutu: Done.

stats/Record.md
bogdandrutu: Done.

stats/Record.md
bogdandrutu: Removed from the moment.

stats/Export.md
bogdandrutu: Not sure I follow your comment here. I think it is wrong in C++ that we combined the View/Aggregation definition with the Data exporting.

stats/DataAggregation.md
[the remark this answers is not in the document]
sebright: I think it would help to make it clearer that this is a comparison with other systems, for example, you could say that tags are similar to labels in some other systems.  Otherwise, it sounds like the document is introducing more OpenCensus terminology.  The same applies to "metric" in other parts of the document.

stats/DataAggregation.md
```diff
@@ -0,0 +1,67 @@
+# Data Aggregation API Overview
+The stats library allows users to define measurements how data are aggregated and broken down.
```
sebright: I'm not sure I understand this sentence.  Should "measurements" be replaced with "view"?

stats/DataAggregation.md
```diff
@@ -0,0 +1,67 @@
...
+The core data types used are:
+* `Aggregation`: defines the aggregation type (e.g. `Distribution`, `Sum`, `Mean`).
+* `View`: defines how individual measurements are broken down by tags and aggregated.
+
+## Aggregation
+An Aggregation describes how data collected is aggregated.
+
+The library SHOULD provide support for multiple types of Aggregations:
+* `Count`: indicates that data collected and aggregated with this `Aggregation` type will be turned
+into a count value.
```
sebright: Something like "Count: counts the number of measurements recorded" might be clearer.

stats/DataAggregation.md
```diff
@@ -0,0 +1,67 @@
...
+A View describes how individual measurements are both broken down (by a set of tag keys, which
+form the basis for a Metric labels.) and aggregated (e.g. into Distributions). A single Measure
+can be used by multiple Views.
+
+A View is defined from the following:
+* `name`: a string by which the View will be referred to, e.g. "rpc_latency". Names MUST be unique
+within the library.
+* `description`: a free format descriptive string, e.g. "RPC counts for last minute and hour"
+* `Measure`: the Measure to which this view is applied.
+* `columns`: an array of tag keys. These values associated with tags of this name and type
```
sebright: Tag keys can no longer have different types.

stats/DataAggregation.md
```diff
@@ -0,0 +1,67 @@
...
+// Define a list of tags to break down view by RPC method.
+ArrayList tagKeyListForMethodViews = new ArrayList();
+tagKeyListForMethodViews.add(RPC_METHOD_KEY);
+
+// Create latency buckets from 100us to 10s.
+BucketBoundaries LATENCY_DISTRIBUTION_BUCKETS =
+    new BucketBoundaries(Arrays.asList(0, 0.01, 0.1, 1.0, 10.0, 1000.0, 10000.0));
+
+// Create a distribution with the latency buckets.
+Distribution LATENCY_DISTRIBUTION = new Distribution.create(LATENCY_DISTRIBUTION_BUCKETS);
```
sebright: I think "new" is redundant here.

stats/Export.md
```diff
@@ -0,0 +1,34 @@
...
+* `MeanData`: data generated for a `Mean` aggregation.
+* `DistributionData`: data generated for a `Distribution` aggregation.
+
+### ViewData
+A ViewData is defined from the following:
+* `name`: the name of the exported `View`.
+* `columns`: the `columns` of the exported `View`.
+* `Measure`: the `Measure` of the exported `View`. This field MAY be replaced with
+`MeasureDescription` if defined.
+* `aggregation`: the `aggregation` of the exported `View`
```
sebright: I think ViewData is supposed to contain the whole View, rather than just the `name`, `columns`, `Measure`, and `aggregation` fields, so that it is completely self-contained (similar to Measure and Measurement).  That is how it is implemented in opencensus-java.

stats/Record.md
```diff
@@ -0,0 +1,68 @@
...
+```
+private static final MeasureDouble RPC_LATENCY =
+    MeasureDouble.create("grpc.io/latency", "latency", "ms");
+```
+
+References to Measures in the system MAY be obtained from querying of registered Measures. This
+functionality is required to decouple the recording of the data from the exporting of the data.
+
+Implementations MAY define a `MeasureDescription` data type which contains all the read-only fields 
+from the `Measure` definition such as: `name`, `description`, `unit` and `type`.
```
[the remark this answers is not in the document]
bogdandrutu: There may be some languages where this is not easy to implement support. It is a suggestion and mostly a name suggestion if this is needed. I added a sentence to clarify. It is only for languages data do not allow private properties/metadata.

stats/README.md
```diff
@@ -0,0 +1,15 @@
...
+This documentation serves to document the "look and feel" of the open source stats API's. It 
+describes the key types and the overall behavior.
+
+## Overview
+OpenCensus allows users to create typed measures, record measurements, define data aggregation, and
+export aggregated data.
+
+### Main APIs
+* [Record API](Record.md): used by the application and third-party libraries (e.g gRPC) owners to
+record metrics.
```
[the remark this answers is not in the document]
bogdandrutu: Removed "third-party".

stats/Record.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Measurement / MeasurementMap
+A `Measurement` describes a value with the following metadata:
+* `Measure`: the `Measure` to which this `value` is applied.
+* `value`: recorded value, MUST have the appropriate type to match the `Measure` definition.
```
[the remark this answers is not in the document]
bogdandrutu: Using fields imply the notion of a class and members. Some languages may not support that.

stats/Record.md
```diff
@@ -0,0 +1,68 @@
...
+
+Implementations MAY define a `MeasurementMap` which describes a set of data points to be collected
+for a set of Measures. Adding this functionality may improve the record usage API.
```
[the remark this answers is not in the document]
bogdandrutu: done.

stats/Record.md
```diff
@@ -0,0 +1,68 @@
...
+current context are recorded with the Measurements if they are any.
+
+Implementations SHOULD provide a means of recording Measurements. This functionality MAY be
+provided through one of the following options:
+* As a method in a class/package (recommended name Stats/stats), taking a list of Measurement as
+argument. e.g. `record(List<Measurement>)` or `record(...Measurement)`.
+* As a `record` method of the appropriate data type. e.g. `MeasurementMap.record()`.
+
+Example in Java
+```
```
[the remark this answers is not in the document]
bogdandrutu: done. don't think we are using but we can set that up.

stats/Record.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Measurement / MeasurementMap
+A `Measurement` describes a value with the following metadata:
+* `Measure`: the `Measure` to which this `value` is applied.
+* `value`: recorded value, MUST have the appropriate type to match the `Measure` definition.
+
+Implementations MAY define a `MeasurementMap` which describes a set of data points to be collected
+for a set of Measures. Adding this functionality may improve the record usage API.
+
+## Record usage
```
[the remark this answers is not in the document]
bogdandrutu: done.

stats/DataAggregation.md
bogdandrutu: Good catch. Fixed.

stats/DataAggregation.md
bogdandrutu: done.

stats/DataAggregation.md
bogdandrutu: done.

stats/DataAggregation.md
bogdandrutu: done.

stats/Record.md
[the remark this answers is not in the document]
bogdandrutu: Rephrase a bit.

stats/Export.md
bogdandrutu: done.

bogdandrutu:
PTAL. I had a hard time with github code review to follow all the comments. If I missed something let me know.

stats/Record.md
```diff
@@ -0,0 +1,68 @@
...
+
+Users should record Measurements with the current context (implicit or explicit). Tags from the
+current context are recorded with the Measurements if they are any.
+
+Implementations SHOULD provide a means of recording Measurements. This functionality MAY be
+provided through one of the following options:
+* As a method in a class/package (recommended name Stats/stats), taking a list of Measurement as
+argument. e.g. `record(List<Measurement>)` or `record(...Measurement)`.
+* As a `record` method of the appropriate data type. e.g. `MeasurementMap.record()`.
```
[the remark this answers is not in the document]
bogdandrutu: Agreed, but I wanted to show more options. Fixed the comment to mention the requirement.

stats/DataAggregation.md
bogdandrutu: I think it makes sense to have some of them required and some optional. We can add that later I think.

stats/Record.md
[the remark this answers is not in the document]
sebright: That is what Github uses to display markdown, but I wasn't sure if we were going to render these pages in other ways.

stats/Record.md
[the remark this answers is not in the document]
sebright: Would those languages' APIs need to expose both Measure and MeasureDescription, or could one be used only in the implementation?

stats/DataAggregation.md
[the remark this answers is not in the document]
sebright: The first sentence in Export.md also uses metrics and labels without mentioning that they are from other projects.

stats/DataAggregation.md
```diff
@@ -0,0 +1,67 @@
...
+private static final BucketBoundaries LATENCY_DISTRIBUTION_BUCKETS =
+    new BucketBoundaries(Arrays.asList(0, 0.01, 0.1, 1.0, 10.0, 1000.0, 10000.0));
+
+// Create a distribution with the latency buckets.
+private static final Distribution LATENCY_DISTRIBUTION =
+    Distribution.create(LATENCY_DISTRIBUTION_BUCKETS);
+
+// Define a view to collect RPC latency stats by RPC method.
+private static final View RPC_LATENCY_BY_METHOD_VIEW = new View(
+  "rpc_latency_by_method", // name
```
songy23: nit: s/"rpc_latency_by_method"/Name.create("rpc_latency_by_method")

stats/Record.md
```diff
@@ -0,0 +1,71 @@
...
+A `Measurement` is defined from the following:
+* `Measure`: the `Measure` to which this `value` is applied.
+* `value`: recorded value, MUST have the appropriate type to match the `Measure` definition.
+
+Implementations MAY define a `MeasurementMap` which describes a set of data points to be collected
+for a set of Measures. Adding this functionality may improve the efficiency of the record usage API.
+
+## Recording Stats
+
+Users should record Measurements with the current context (implicit or explicit). Tags from the
```
songy23: > [For that matter, there is no real concept of "current explicit context". I would favor "Users record Measurements against a context, either an explicit context or the implicit current context."](https://github.com/census-instrumentation/opencensus-specs/pull/51/commits/c632ef9817b432e6eea1693d4bc7b8bdfb2c50de#diff-d00b38026435b3cd635ac7469f875fb4R53)

Please fix.

bogdandrutu: done.

stats/DataAggregation.md
bogdandrutu: done.

stats/DataAggregation.md
```diff
@@ -0,0 +1,71 @@
...
+An Aggregation describes how data collected is aggregated.
+
+The library SHOULD provide support for multiple types of Aggregations:
+* `Count`: counts the number of measurements recorded.
+* `Sum`: indicates that data collected and aggregated with this `Aggregation` will be summed up.
+* `Mean`: indicates that data collected and aggregated with this `Aggregation` will calculate the
+mean value.
+* `Max`: indicates that data collected and aggregated with this `Aggregation` will calculate the 
+maximum value recorded.
+* `LastValue`: indicates that data collected and aggregated with this `Aggregation` will
```
adriancole: fyi @jkschneider there are a couple new things here as well an explanation of views if you have some time to comment.

stats/DataAggregation.md
```diff
@@ -0,0 +1,71 @@
+# Data Aggregation API Overview
+The stats library allows users to define views to configure how data are aggregated and broken down.
+The core data types used are:
+* `Aggregation`: defines the aggregation type (e.g. `Distribution`, `Sum`, `Mean`).
+* `View`: defines how individual measurements are broken down by tags (similar to labels in some 
+other systems) and aggregated.
+
+## Aggregation
```
jkschneider: Client-side percentiles are also a very common ask when the monitoring system doesn't support aggregable percentile derivation from histogram data (which I assume is the purpose of `Distribution`?). They don't seem to fit in one of these types.

stats/DataAggregation.md
```diff
@@ -0,0 +1,71 @@
...
+An Aggregation describes how data collected is aggregated.
+
+The library SHOULD provide support for multiple types of Aggregations:
+* `Count`: counts the number of measurements recorded.
+* `Sum`: indicates that data collected and aggregated with this `Aggregation` will be summed up.
+* `Mean`: indicates that data collected and aggregated with this `Aggregation` will calculate the
+mean value.
+* `Max`: indicates that data collected and aggregated with this `Aggregation` will calculate the
```
jkschneider: Can we somehow indicate that this needs to be a decaying value to be useful?

stats/DataAggregation.md
```diff
@@ -0,0 +1,71 @@
...
+* `View`: defines how individual measurements are broken down by tags (similar to labels in some 
+other systems) and aggregated.
+
+## Aggregation
+An Aggregation describes how data collected is aggregated.
+
+The library SHOULD provide support for multiple types of Aggregations:
+* `Count`: counts the number of measurements recorded.
+* `Sum`: indicates that data collected and aggregated with this `Aggregation` will be summed up.
+* `Mean`: indicates that data collected and aggregated with this `Aggregation` will calculate the
```
jkschneider: I worry about this one. `Mean` isn't aggregable across dimensions, so doesn't seem particularly useful in general. For distribution summary type metrics, shipping a total amount and count is generally sufficient to be able to derive a truly aggregable mean. Some systems are so aware of this problem that they provide a standard query function OOTB (see [dist-avg](https://github.com/Netflix/atlas/wiki/math-dist%E2%80%90avg). Some monitoring systems expect mean to be pre-computed as a matter of convention, but could shipping mean be an exporter concern for those systems?

deadtrickster: easy calculate with Prometheus too.

stats/Record.md
```diff
@@ -0,0 +1,71 @@
...
+functionality is required to decouple the recording of the data from the exporting of the data.
+
+For languages that do not allow private properties/metadata and if they are needed implementations 
+MAY define a `MeasureDescription` data type which contains all the read-only fields  from the 
+`Measure` definition such as: `name`, `description`, `unit` and `type`.
+
+## Measurement
+A `Measurement` is defined from the following:
+* `Measure`: the `Measure` to which this `value` is applied.
+* `value`: recorded value, MUST have the appropriate type to match the `Measure` definition.
```
deadtrickster: Can someone give an example of appropriate / not appropriate values?

stats/Record.md
```diff
@@ -0,0 +1,71 @@
...
+includes information such as the type of measurement, the units of measurement and descriptive names
+for the data. This provides the fundamental type used for recording data.
+
+A Measure describes a value with the following metadata:
+* `name`: a string by which the measure will be referred to, e.g. "rpc_server_latency", or
+"vm_cpu_cycles". Names MUST be unique within the library. It is recommended to use names 
+compatible with the intended end usage, e.g, use host/path pattern.
+* `description`: a string describes the measure, e.g. "RPC latency in seconds", "Virtual cycles
+executed on VM".
+* `unit`: a string describing the unit used for the `Measure`. Follows the format described by
```
deadtrickster: below unit referenced in value validation context, is it also exported somehow? what views know about units?

stats/DataAggregation.md
```diff
@@ -13,6 +13,10 @@ The library SHOULD provide support for multiple types of Aggregations:
 * `Sum`: indicates that data collected and aggregated with this `Aggregation` will be summed up.
 * `Mean`: indicates that data collected and aggregated with this `Aggregation` will calculate the
 mean value.
+* `Max`: indicates that data collected and aggregated with this `Aggregation` will calculate the
```
songy23: Stackdiver Monitoring has a concept [Range](https://cloud.google.com/monitoring/api/ref_v3/rest/v3/TypedValue#Range), which includes both max and min values. I think using a similar `Range` aggregation probably makes more sense.

stats/DataAggregation.md
bogdandrutu: Thanks Adrian.

stats/DataAggregation.md
bogdandrutu: I filed an issue for this https://github.com/census-instrumentation/opencensus-specs/issues/55. I will address soon.

stats/DataAggregation.md
bogdandrutu: Drop this for the moment. Added an issue to discuss the need for this https://github.com/census-instrumentation/opencensus-specs/issues/56

stats/Record.md
bogdandrutu: Yes in the exported data we will include this. View do not use this unit but backends may use this to display in the UI or to apply data conversion (we record KB for each measurement but when displayed the backend may display MB/sec for example).

stats/Record.md
bogdandrutu: For example if you have an int64 measure, recording the value 2.3 is invalid. This cannot be achieved in all the languages.

stats/DataAggregation.md
bogdandrutu: How would you phrase that?

bogdandrutu:
I think it is important for the project to have this merged, but comments are very important as well. People who want to comment more about this can open issues or PRs after the merge or even still comment on this PR.

Thanks.

PR #52, opened 2018-02-26
Add http/path and http/route Stackdriver Trace labels

rakyll:
Stackdriver will support path and route as a first-class label in the UI.

As well, as use the new short label keys rather than the
fully namespaced ones for HTTP.

### DIFF
--- a/trace/HTTPAttributes.md
+++ b/trace/HTTPAttributes.md
@@ -25,14 +25,14 @@ known attributes/labels.
 The following table summarizes how OpenCensus attributes maps to the
 known attributes/labels on supported tracing backends.
 
-| OpenCensus attribute      | Zipkin             | Jaeger             | Stackdriver Trace label                    |
-|---------------------------|--------------------|--------------------|--------------------------------------------|
-| "http.host"               | "http.host"        |                    | "trace.cloud.google.com/http/host"         |
-| "http.method"             | "http.method"      | "http.method"      | "trace.cloud.google.com/http/method"       |
-| "http.path"               | "http.path"        |                    |                                            |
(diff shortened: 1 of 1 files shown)


trace/HTTPAttributes.md
```diff
@@ -25,14 +25,14 @@ known attributes/labels.
...
-|---------------------------|--------------------|--------------------|--------------------------------------------|
-| "http.host"               | "http.host"        |                    | "trace.cloud.google.com/http/host"         |
-| "http.method"             | "http.method"      | "http.method"      | "trace.cloud.google.com/http/method"       |
-| "http.path"               | "http.path"        |                    |                                            |
-| "http.route"              | "http.route"       |                    |                                            |
-| "http.user_agent"         |                    |                    | "trace.cloud.google.com/http/user_agent"   |
-| "http.status_code"        | "http.status_code" | "http.status_code" | "trace.cloud.google.com/http/status_code"  |
+| OpenCensus attribute      | Zipkin             | Jaeger             | Stackdriver Trace label  |
+|---------------------------|--------------------|--------------------|--------------------------|
+| "http.host"               | "http.host"        |                    | "http/host"              |
```
draffensperger: Minor nit: we have typically prefixed the Stackdriver Trace labels with a preceding "/", so they would be "/http/host", etc.

That's how for instance the pre-Census Stackdriver Trace Node.js agent does it, see: https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/blob/26d6f59236abcf5c4beaba2c7dff69ac2c1479fe/src/trace-labels.ts#L27

rakyll: Done

PR #60, opened 2018-03-02
Add a spec for HTTP tracing

rakyll:
Merge the HTTPAttributes spec into the new one.


tsloughter:
For the message events I assume it is expected to record both response and request sizes on both the client and server spans?

trace/HTTP.md
```diff
@@ -0,0 +1,105 @@
...
+This document explains tracing of HTTP requests with OpenCensus.
+
+## Spans
+
+Implementations MUST create a span for outgoing requests at the
+client and a span for incoming requests at the server.
+
+Span name is formatted as:
+
+* Sent.$host[:$port]/$path for outgoing requests.
```
HailongWen: We use "Sent.className.methodName" in current java instrumentation. Should we change to follow this new convention? /cc @bogdandrutu

trace/HTTP.md
```diff
@@ -0,0 +1,105 @@
...
+If user doesn't set any propagation methods explicitly, B3 is used.
+
+The propagation method SHOULD modify a request object to insert
+a SpanContext or SHOULD be able to extract a SpanContext from a
+request object.
+
+## Status
+
+Implementations should set status if HTTP request or response
+is not successful (not 2xx).
```
HailongWen: what about 3xx if the HTTP client does not have an redirect mechanism, is that still counted as successful?

rakyll: IMHO, a 3xx should be considered successful in my opinion. Because it satisfies the redirect request. The user can explicitly set an error status if they are not satisfied with an HTTP 3xx. I will add a note about3xx here.

trace/HTTP.md
rakyll: @bogdandrutu agrees that we should use this naming format.

"Sent.className.methodName" should always be the same so it doesn't help identifying the span.

adriancole:
>
> +Span name is formatted as:
> +
> +* Sent.$host[:$port]/$path for outgoing requests.
>
> feedback and conventions on zipkin include the http method fyi. span name
in zipkin is low cardinality and I don't think including the host will help
with this. So when mapping in zipkin we can't use this as a span name (it
will make the span name dropdown absurd in some sites)

> @bogdandrutu <https://github.com/bogdandrutu> agrees that we should use
> this naming format.
>
> "Sent.className.methodName" should always be the same so it doesn't help
> identifying the span.
>
also this is http abstraction, we don't know if there is a mapping to a
model layer (even if usually there is). the controller doing the send could
be an attribute in another namespace or a parent span.

trace/HTTP.md
HailongWen: Ack.

rakyll:
PTAL

trace/HTTP.md
```diff
@@ -0,0 +1,114 @@
...
+
+Set status code to UNKNOWN (2) if the reason cannot be inferred
+at the callsite or from the HTTP status code.
+
+Don't set the status message if the reason can be inferred at
+the callsite of from the HTTP status code.
+
+## Message events
+
+In the lifetime of an incoming and outgoing request, the following
```
adriancole: this is easier to do from low-level libraries. should we mention explicitly chunked encoding or if it is satisfactory to use Content-Length?

trace/HTTP.md
```diff
@@ -0,0 +1,114 @@
...
+|---------------------------|--------------------|--------------------|---------------------------|
+| "http.host"               | "http.host"        | *                  | "/http/host"              |
+| "http.method"             | "http.method"      | "http.method"      | "/http/method"            |
+| "http.path"               | "http.path"        | *                  | "/http/path"              |
+| "http.route"              | "http.route"       |                    | "/http/route"             |
+| "http.user_agent"         |                    |                    | "/http/user_agent"        |
+| "http.status_code"        | "http.status_code" | "http.status_code" | "/http/status_code"       |
+
+
+(*) Jager supports "http.url" instead of a path and method attribute.
```
adriancole: actually this policy was likely inherited from Zipkin back before uber forked it https://github.com/openzipkin/zipkin/commit/539a08c19a013eb9380aa8c040a97780bf77de63

trace/HTTP.md
```diff
@@ -0,0 +1,114 @@
...
+OpenCensus doesn't collect the full URL for security reasons.
```
adriancole: collect is local to the process.. I think we'll need to distinguish the impact of this later, but fine for now

trace/HTTP.md
```diff
@@ -0,0 +1,114 @@
...
+Exporters can build a URL from the collected method and path and
```
adriancole: this will work for jaeger assuming they've defined it w/o query params and you can infer the scheme. This won't work for zipkin

trace/HTTP.md
yurishkuro: Jaeger doesn't do anything special with any of these tags, so "supports" is a strong/misleading word, a user can always send any tag, including all the tags listed in the "OpenCensus attribute" column.

rakyll: @yurishkuro, if Jaeger doesn't have any additional current or planned UI features for any of these tags, I should probably document OpenCensus tags as the tags to be used when exporting to Jaeger.

We do the mapping if the backend backend has a specific feature for the tag.

yurishkuro: yes, the OC tags can be copied as is when reporting to Jaeger, especially since they match the OT tags where available.

rakyll: Done.

rakyll: I removed this section and suggested using OpenCensus tags as they are when there is not a match.

trace/HTTP.md
rakyll: I think we need to define behavior for chunked and multipart requests. I will file an issue and follow-up with a proposal.

trace/HTTP.md
```diff
@@ -0,0 +1,109 @@
...
+-> [time], MessageEventTypeRecv, UncompressedByteSize, CompressedByteSize
+```
+
+## Attributes
+
+Implementations SHOULD set the following attributes on the client
+and server spans. For a server, request represents the incoming request.
+For a client, request represents the outgoing request.
+
+All attributes are optional.
```
odeke-em: Contrary to the opening sentence `Implementations SHOULD set the following attributes on the client and server spans`

adriancole:
Thanks!

Issue #45, opened 2018-01-22
Selective Tag propagation


ramonza:
When instrumenting a library, you want to be as generous as possible with tags that you provide so that the Operator (whoever ends up using your library in an application) has maximum flexibility in how they construct views to break down stats.

However, currently all tags are propagated (for example in gRPC) by default. This means that there's a cost added for every new tag, a cost that is only visible to the Instrumentor (person instrumenting the library) but really paid by the Operator (application developer).

To solve this, we need a way for the Operator to limit tag propagation.

ramonza:
Initial idea discussed with @bogdandrutu is by default *not* propagate any tags and then to provide an API similar to Views to enable their propagation, probably under a different name.

For example, you may want to propagate the gRPC service name tag under a different tag name (so that it doesn't conflict with downstream services). 

Under this model, there's no cost to adding more tags if they aren't propagated and they aren't used in views.

adriancole:
Another way of saying this is that there's a whitelist approach, right? IOTW tags exist in a namespace, regardless, and we are filtering some to propagate to the next hop, subject to a potential rename policy.

One interesting bit about this is that tags currently are validated and constrained constants, due to downstream propagation limitations. Not all of these limitations apply to in-process tags, particularly if you are inheriting a separate set of tags from a different context. Just something to think about when we consider teasing things out. If we aggressively up-front validate for remote propagation, we may end up dropping valid tags for correlation

SergeyKanzhelev:
Should there be separate collection for `Trace-Context` tags? Those are quite special as you want to allow tracing libraries modify them, but not expose to end user of the tracing library.

For the rest of the tags - one way to limit those may be set `TTL=0` on a tag. Similar to [TTL=1](https://github.com/w3c/distributed-tracing/issues/23) here.

ramonza:
>Another way of saying this is that there's a whitelist approach, right? IOTW tags exist in a namespace, regardless, and we are filtering some to propagate to the next hop, subject to a potential rename policy.

Yes, I think this is functionally equivalent to a whitelist although the name "whitelist" doesn't feel right to me, maybe because I think of it in the same was as the distinction between Measures/Views.

[...]

PR #69, opened 2018-03-14
Add a Scope property on TagKey

ramonza:
This is a start of #45

### DIFF
--- a/tags/TagContext.md
+++ b/tags/TagContext.md
@@ -13,10 +13,22 @@ all of the information that must be propagated across process boundaries.
 
 ## TagKey
 
-A string or string wrapper, with some restrictions:
+A TagKey consists of a Name and a Scope.
+
+The Name of a TagKey is a string or string wrapper, with some restrictions:
 
 - Must contain only printable ASCII (codes between 32 and 126, inclusive).
 - Must have length greater than zero and less than 256.
+
+Scope indicates the default propagation behavior. The Scope may be one of:
+
+* Local: Tags with this key are local to the current process and are not
+  propagated across process boundaries, but may be propagated within-process.
+* Request: Tags with this key are scoped to the processing of this request,
+  and are propagated across process boundaries.
+
+Scope should usually be explicitly specified. In cases where this isn't 
+convenient, the default scope should be Local.
 
 ## TagValue
 


tags/TagContext.md
```diff
@@ -13,11 +13,22 @@ all of the information that must be propagated across process boundaries.
...
+The Name of a TagKey is a string or string wrapper, with some restrictions:
 
 - Must contain only printable ASCII (codes between 32 and 126, inclusive).
 - Must have length greater than zero and less than 256.
 
+Scope indicates the default propagation behavior. The Scope may be one of:
+
+* Local: Tags with this key are local to the current process and are not
+  propagated across process boundaries, but may be propagated within-process.
+* Trace: Tags with this key are scoped to the current distributed trace,
```
ramonza: spoke to @bogdandrutu we are going to rename this to "request scoped" in order to not imply a relationship to tracing

tags/TagContext.md
```diff
@@ -13,11 +13,23 @@ all of the information that must be propagated across process boundaries.
 
 ## TagKey
 
-A string or string wrapper, with some restrictions:
+A TagKey consists of a Name and a Scope.
```
sebright: I would rather not specify that the scope is an attribute of TagKey.  It may make sense to support tag scope in some implementations by providing a way to "register" a tag for propagation, with a function that takes a TagKey and a Scope.  For example, it could work like this in Java:

```java
static final TagKey MY_KEY = ...
static final TagManager tagManager = ...

static initialize() {
  ...
  tagManager.registerTagScope(MY_KEY, TagScope.REQUEST);
```

This API is more flexible because it separates creating tag keys from setting their scope, and it avoids adding scope information to views, which contain TagKeys but don't need to specify how the tags are propagated.

adriancole:
Interesting as I have other works similar. Ex a whitelist for propagated
keys. Is there a local scope? Ex tag that doesnt leave the process but is
propagated within a request up till the point of exit? (No farther).
Usually but not always these tags are per process like instance ID, but
some are not.

ramonza:
I'm going to write up a larger proposal for this. I just wanted to kick this off since it might impact our API stability goals if we need to make a change here.

tags/TagContext.md
adriancole: fyi there's some discussion here.. I do like the idea of "request" and "local" scopes https://github.com/openzipkin/brave/pull/577

Also, I'm not sure, but it does feel like a registration approach could be helpful. I wonder if we shouldn't start with an experimental api before doing a spec. At least before merging a spec

Issue #53, opened 2018-02-27
Should tag values be less restrictive?


ramonza:
Tag values as specified in tags/TagContext.md are as restrictive as keys:

> 
> TagKey
> A string or string wrapper, with some restrictions:
> 
> Must contain only printable ASCII (codes between 32 and 126, inclusive).
> Must have length greater than zero and less than 256.
> TagValue
> A string or string wrapper with the same restrictions as TagKey, except that it is allowed to be empty.
> 

This seems too restrictive. Common metrics backends seem to support arbitrary values:
* https://prometheus.io/docs/concepts/data_model/

Also I would guess that it would be very inconvenient for non-latin languages where you could imagine things like city names being in tag values (for example). I would be very frustrated if the backend I'm using supports my native language characters but I'm forced to encode everything just because OpenCensus creates an artificial restriction.

adriancole:
I suppose this depends on if the intended transport for tags will be easier
for values vs keys. But more importantly tags are propagated, and I'm not
sure there's much prior art on this except a myriad of problems due to
format or size (encoding onto http requests) :P

I think correlation fields certainly can be more flexible than the subset
of which that are propagated. Also not sure if the format isn't too
restrictive, just mentioning the difference.

songy23:
Obsolete.

PR #71, opened 2018-03-15
Allow any unicode characters in the TagValue

ramonza:
Prometheus permits any characters:
https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels

In gRPC propagation, the tags are binary encoded as UTF-8 strings,
so there should be no problem there. For HTTP, we can use mime
encoding.

### DIFF
--- a/tags/TagContext.md
+++ b/tags/TagContext.md
@@ -20,8 +20,8 @@ A string or string wrapper, with some restrictions:
 
 ## TagValue
 
-A string or string wrapper with the same restrictions as `TagKey`, except that it
-is allowed to be empty.
+A TagValue can be any unicode string whose size when encoded in UTF-8
+is less than 256 bytes. An empty value is allowed.
 
 ## Serialization
 


ramonza:
I feel that we are being unreasonably US-centric in restricting tag values to latin characters. It is very easy to imagine useful tags that contain non-latin character, for example the names of Chinese metro areas. We should not force people to latinize all these just to use OpenCensus.

bogdandrutu:
Fixes https://github.com/census-instrumentation/opencensus-specs/issues/53.

bogdandrutu:
@ramonza I would like to know if the set of characters that you propose here are accepted by few of the major backends that we try to support:
* Prometheus - which does
* SignalFx - ???
* Stackdriver - ???

dinooliva:
Pragmatically, we didn't allow this because of compatibility issues internal to google and because it allows fast encoding/decoding to languages that don't natively use UTF8 (e.g. Java).

More generally, we've had discussions about making TagValue an uninterpreted sequence of bytes, which probably makes more sense. Making this change requires some rollout effort so better to design the solution and do it once.

ramonza:
@dinooliva I think it would make for very strange APIs for tag values to be uninterpreted bytes. perhaps we should just remove the statement about length limit and instead of it being an error allow the user to truncate the values at the point of serializing them (when it really matters)? cc @bogdandrutu

Issue #84, opened 2018-03-28
Document the default sampling rate for tracing


rakyll:
Default tracing rate (1e-4) is not documented.

PR #87, opened 2018-03-29
Define the default sampler.

bogdandrutu:
Fixes https://github.com/census-instrumentation/opencensus-specs/issues/84

### DIFF
--- a/trace/TraceConfig.md
+++ b/trace/TraceConfig.md
@@ -5,7 +5,8 @@ sampler, maximum events to be kept, etc.
 
 ## TraceParams
 Represents the set of parameters that users can control 
-* Default `Sampler` - used when creating a Span if no specific sampler is given.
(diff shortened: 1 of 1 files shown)


trace/TraceConfig.md
```diff
@@ -5,7 +5,8 @@ sampler, maximum events to be kept, etc.
 
 ## TraceParams
 Represents the set of parameters that users can control 
-* Default `Sampler` - used when creating a Span if no specific sampler is given.
+* Default `Sampler` - used when creating a Span if no specific sampler is given. The default sampler
```
ramonza: I worry that we will want to replace this with a rate limited sampler as the default once this is available. For most applications, rate limited seems to be a better default. Maybe we should make a note that we might replace it in future with a rate limit but it will stil not exceed 1/10000 by default.

PR #88, opened 2018-03-29
Remove '.' from the tag names in grpc and http specs.

bogdandrutu:
This is because the main stats backends that we currently support Prometheus and Stackdriver they do not support '.':
https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels

Unfortunately I cannot find the SD requirements, but I manually tested and '.' is not accepted.

This does not mean for the moment that we force this rule for all the tags, but we do it here for the most common tags at least. We will re-evaluate what is the best approach for sanitization later.

### DIFF
--- a/stats/HTTP.md
+++ b/stats/HTTP.md
@@ -36,10 +36,10 @@ All client metrics should be tagged with the following.
 
 | Tag suffix         | Description                                                                                              |
 |--------------------|----------------------------------------------------------------------------------------------------------|
-| http.client_method | HTTP method, capitalized (i.e. GET, POST, PUT, DELETE, etc.)                                             |
-| http.client_path   | URL path (not including query string)                                                                    |
-| http.client_status | HTTP status code as an integer (e.g. 200, 404, 500.), or "error" if no response status line was received |
-| http.client_host   | Value of the request Host header                                                                         |
+| http_client_method | HTTP method, capitalized (i.e. GET, POST, PUT, DELETE, etc.)                                             |
+| http_client_path   | URL path (not including query string)                                                                    |
+| http_client_status | HTTP status code as an integer (e.g. 200, 404, 500.), or "error" if no response status line was received |
+| http_client_host   | Value of the request Host header                                                                         |
 
 ### Default views
 
@@ -74,10 +74,10 @@ All server metrics should be tagged with the following.
 
 | Tag suffix         | Description                                                         |
 |--------------------|---------------------------------------------------------------------|
-| http.server_method | HTTP method, capitalized (i.e. GET, POST, PUT, DELETE, etc.)        |
-| http.server_path   | URL path (not including query string)                               |
-| http.server_status | HTTP server status code returned, as an integer e.g. 200, 404, 500. |
-| http.server_host   | Value of the request Host header                                    |
+| http_server_method | HTTP method, capitalized (i.e. GET, POST, PUT, DELETE, etc.)        |
+| http_server_path   | URL path (not including query string)                               |
+| http_server_status | HTTP server status code returned, as an integer e.g. 200, 404, 500. |
+| http_server_host   | Value of the request Host header                                    |
 
 ### Default views
 
--- a/stats/HTTP.md
+++ b/stats/HTTP.md
@@ -69,8 +69,8 @@ Server stats are recorded for each individual HTTP request, including for each r
 ### Tags
 
 All server metrics should be tagged with the following. 
-`http.server_method`, `http.server_path`, `http.server_host` are available in the context for the entire request processing.
-`http.server_status` is only available around the stats recorded at the end of request processing.
+`http_server_method`, `http_server_path`, `http_server_host` are available in the context for the entire request processing.
+`http_server_status` is only available around the stats recorded at the end of request processing.
 
 | Tag suffix         | Description                                                         |
 |--------------------|---------------------------------------------------------------------|
--- a/stats/gRPC.md
+++ b/stats/gRPC.md
@@ -89,8 +89,8 @@ All server metrics should be tagged with the following.
 | grpc_server_method | Full gRPC method name, including package, service and method, e.g. com.exampleapi.v4.BookshelfService/Checkout |
 | grpc_server_status | gRPC server status code returned, e.g. OK, CANCELLED, DEADLINE_EXCEEDED                                        |
 
-`grpc.server_method` is available in the context for the entire RPC call handling. 
-`grpc.server_status` is only available around metrics recorded at the end of the request.
+`grpc_server_method` is available in the context for the entire RPC call handling. 
+`grpc_server_status` is only available around metrics recorded at the end of the request.
 
 ### Default views
 
--- a/stats/gRPC.md
+++ b/stats/gRPC.md
@@ -44,8 +44,8 @@ All client metrics should be tagged with the following.
 
 | Tag name           | Description                                                                                                      |
 |--------------------|------------------------------------------------------------------------------------------------------------------|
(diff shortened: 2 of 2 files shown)


adriancole:
This is a big change, though looks small. Whats the rationale?

ramonza:
@adriancole some important backends don't support "." in labels, e.g. https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels

adriancole:
Im aware of this, but you should know you are throwing another under the
bus which is years of convention in zipkin ecosystem.

When you say "some important backends" is it plural or do you just mean
prometheus?

ramonza:
@adriancole these tags are not added to traces - they are purely for aggregating stats so I don't see how consistency with Zipkin would come into the picture?

Regarding other backends, Stackdriver also does not support periods in label names either.

I believe we're proposing this change under the assumption that changing it now has virtually no downside (since we're changed other properties of the view this week) and so we may as well make the tags as convenient to use as possible. 
If that's not the case and there really is some important downside here, we could consider alternatives to this such as sanitizing in the exporter?

adriancole:
@ramonza now you are getting to what I was looking for, which is rationale :)

Let's try to put rationale on issues like this.. there is definitely downside because solidifying conventions with only a couple backends won't be in consideration with them.

So, here are some rationale (probably good to keep rationale.md or something)

* we assume people aren't cross-querying or won't be able to help with that if they do
  * for example, in systems where you query across traces and metrics by the same label, this won't help with that *even in systems that have compatible conventions*.
* we assume that in permitting stackdriver and prometheus with _ we aren't excluding systems who might not support underscore (no idea if that's true, but glad we are talking concrete things)

So anyway, you've answered my inquiry which was some rationale and what we are optimizing for, what knowledge we have when we do it. That's the most important thing, which is in a spec repo to vocalize (for lack of better term) what tradeoffs we are making and with what information.

Make sense?

jkschneider:
> some important backends don't support "." in labels, e.g. 

It starts with labels, but when we consider the bigger picture that includes metric names as well, the insufficiency of picking a single naming convention becomes more apparent. For example, SignalFx almost requires names as name separators in metric names because the UI uses it for grouping. Dots in metric names in Elastic break the indexing mechanism. 

Without some naming convention normalization mechanism that is peculiar to each implementation, the chosen opinion will eventually run afoul of some monitoring system.

ramonza:
@adriancole you're totally right, we should have added rationale. I kind of like PEP-style process but might be too heavyweight for now. Maybe we could just have a very lightweight template for commit messages to follow that outlines a requirement for rationale?

On this particular PR, here's why I thought we should go ahead with this change:

1. These tags were defined a few days days ago in afbb08857d94a7876ff69ec0b34e596da72f5c49
2. AFAIK the tag keys were arbitrarily chosen strings, they don't relate to anything else in any other backend or any other system, they were just picked to be reasonably descriptive. It sounds like this might be an incorrect assumption. If so, please tell me and if you can it would be super helpful to get a pointer to where they were chosen from or let's get that reference added to the markdown file. 
3. The previous values are not implemented in any OpenCensus library so changing this doesn't affect any working code
4. The arbitrary values that were chosen were found to be inconvenient for two supported metrics backends we currently support

So based on my assumptions, the costs of making this change now were really small and the benefits is that we reduce friction for a large number of users. 

>  because solidifying conventions with only a couple backends won't be in consideration of those not added yet.

Definitely. But this change only strictly increases the compatibility because the previous keys already contained an underscore character. So we aren't excluding any future systems by making this change.

In summary: you're absolutely right that we need to add rationale as a matter of course to all changes on this repo. Thank you for taking the time to explain this and encouraging this good practice :)

@jkschneider you make a great point with SignalFX requiring for grouping. I absolutely think that we need to support transforming metric names and probably also tag keys/labels on the way to the backend. I see metric/view names and tag keys as slightly different though: it seems like a common assumption that tag keys are single "words" while metric names are hierarchical. Yeah there is some hierarchy implicit in the prefixes but fundamentally I don't see as much of a case for making them very structured as I do for metric names (used for hierarchical grouping and such). 

I don't think we should preclude sanitizing tag keys. I do think that judiciously choosing tag keys to reduce the likelihood of needing sanitization for very commonly used tags is reasonable. I also think we need a mechanism to rename tags at the view layer which we currently don't have so that the user can choose their own names that work for their particular situation.

ramonza:
@rakyll I think we need to keep a difference between trace attributes and tags. Tag are propagated on the wire for every RPC potentially so it's important that we keep them concise. Trace attributes are heavily sampled so we can expect people to aggressively add as much debugging information as they need - so in this case it's much more important that they are properly namespaced.

By adding a period, we strayed from our previous convention that tag keys were basically "words" - they had no explicit structure - they match "\w+".  This is just a convention not a requirement I'm suggesting but I think it will help not to bloat propagation headers, especially for commonly used tags.

stats/HTTP.md
```diff
@@ -36,10 +36,10 @@ All client metrics should be tagged with the following.
 
 | Tag suffix         | Description                                                                                              |
 |--------------------|----------------------------------------------------------------------------------------------------------|
-| http.client_method | HTTP method, capitalized (i.e. GET, POST, PUT, DELETE, etc.)                                             |
```
rakyll: Why do we have separate tags for clients and servers? Client measurements will never be aggregated on a server views and vice versa.

stats/gRPC.md
```diff
@@ -86,8 +86,8 @@ All server metrics should be tagged with the following.
 
 | Tag name           | Description                                                                                                    |
 |--------------------|----------------------------------------------------------------------------------------------------------------|
-| grpc.server_method | Full gRPC method name, including package, service and method, e.g. com.exampleapi.v4.BookshelfService/Checkout |
-| grpc.server_status | gRPC server status code returned, e.g. OK, CANCELLED, DEADLINE_EXCEEDED                                        |
+| grpc_server_method | Full gRPC method name, including package, service and method, e.g. com.exampleapi.v4.BookshelfService/Checkout |
+| grpc_server_status | gRPC server status code returned, e.g. OK, CANCELLED, DEADLINE_EXCEEDED                                        |
 
 `grpc.server_method` is available in the context for the entire RPC call handling.
```
adriancole: nit: if all underscores is the way (which may be fine), you'll want to global replace.  I made the same mistake recently in another project and someone had to come back with another commit.

adriancole:
Hi, Ramon. I think we don't need to get to PEP or anything near that, just normal pull requests. For example, no change to spec lacking a description. If changing something based on experience note what experiences drive you. That even lacking a rationale.md (which is used in the trace context spec) is a good start.

On the tags vs attributes.. I personally did think that these were attributes as it looked like a list of dimensions, then discussion further was about storage concerns (not propagation ones which would be format drivers). When I switched mind to this being about propagation, some tag, notably client http status really jumped out at me.. how would that work? We don't have the response until after we've propagated the request. are we talking upstream propagation, or is this just a copy/paste thing?

One of the concerns I've had with other projects is that specs just have a list of things with no guidance on what they are used for, or how to use them, or what they combine with to be useful. Maybe not in this change, but the client status tag shocked me and made me realise there's an opportunity to clean this up. Other spec projects out there have made things worse because defining things before we know how they would be used definitely has caused a large amount of tech debt.

ramonza:
Actually the trace attributes are more or less the same size as tags, I was under the mistaken impression they were namespaces with domain name. So I retract my previous comment +1 for making them the same as trace attributes as suggested by @rakyll

stats/gRPC.md
bogdandrutu: done.

stats/HTTP.md
bogdandrutu: See in the grpc.md the FAQ section. I don't know where to put that section, maybe we should move it into a faq.md.

PR #89, opened 2018-04-06
Revert "Allow any unicode characters in the TagValue #71"

songy23:
Reverts https://github.com/census-instrumentation/opencensus-specs/pull/71.

Based on our meeting today, we need more discussions on supporting Unicode.


rakyll:
Also reverted the Go change, see https://github.com/census-instrumentation/opencensus-go/commit/5e4356a1b558943ae8f9a60319c2bea99e99480c.

PR #97, opened 2018-04-10
Add explicit reference for status code strings

ramonza:


stats/gRPC.md
```diff
@@ -51,6 +51,8 @@ All client metrics should be tagged with the following.
 metrics. 
 `grpc_client_status` is set when an outgoing request finishes and is only available around metrics
 recorded at the end of the outgoing request.
+Status codes should be stringified according to:
+https://github.com/grpc/grpc/blob/ee6e4dc/doc/statuscodes.md
```
sebright: Is there a way to link to the latest status code documentation?  That might be better than a specific commit, since the description could be updated.

ramonza: Pointed to master

stats/gRPC.md
```diff
@@ -96,8 +98,11 @@ All server metrics should be tagged with the following.
 
 `grpc_server_method` is set when an incoming request starts and is available in the context for
 the entire RPC call handling. 
+
 `grpc_server_status` is set when an incoming request finishes and is only available around metrics
 recorded at the end of the incoming request.
+Status codes should be stringified according to:
+https://github.com/grpc/grpc/blob/master/doc/statuscodes.md
```
bogdandrutu: I would use this link:
https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto

ramonza: I considered that, but since this document is all about gRPC I think it's better to point to gRPC docs.

bogdandrutu: I am fine either way.

Issue #66, opened 2018-03-13
tags: implement spec for gRPC propagation


odeke-em:
We lack a spec for tags propagation using gRPC although a couple of the packages already have it, but we've found some issues like in the Go implementation https://github.com/census-instrumentation/opencensus-go/issues/541 where "method" was inserted during propagation.

Also this came up in the equivalent HTTP propagation specs request issue  https://github.com/census-instrumentation/opencensus-specs/issues/65

bogdandrutu:
@ramonza this will be covered by your doc related to tag scopes correct?

songy23:
I believe this can be marked as resolved with #233. Feel free to reopen if there're further questions/comments.

PR #104, opened 2018-04-17
Add document about what to do with untrusted requests.

bogdandrutu:
This is an early draft and I would like to get feedback, we need to do the following:
* decide how do we allow users to control the sampler for untrusted requests (see the comment in the text about this)
* decide how the tags filter should look like, but that will happen when solving this https://github.com/census-instrumentation/opencensus-specs/issues/66
* decide what configs we support in the plugins (grpc, http).

### DIFF
--- a/utils/HandleUntrustedRequests.md
+++ b/utils/HandleUntrustedRequests.md
@@ -1,0 +1,52 @@
+# Handle Untrusted Requests
+
+This document is about how to handle untrusted requests for trace and tags.
+
+## What is an untrusted request?
+An untrusted request refers to a request coming from an external untrusted source (e.g. external
+customer request).
+
+## How does OpenCensus know if a request is trusted or not?
+The OpenCensus library does not have a mechanism to determine if an incoming request is trusted 
+or not, because of this the application owner should configure the library accordingly.
+
+The OpenCensus library should allow to configure this information in the server plugins (grpc, 
+http) for each server instance.
+
+## Why not trusting all the requests?
+For trace we do not trust the incoming requests (trace headers) because:
+* Users can send always sampled requests all the time (by mistake, or malicious user).
+* Users can send always the same trace id (e.g. malicious user).
+
+For tags we do not trust the incoming requests (tags headers) because:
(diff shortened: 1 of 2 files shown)


basvanbeek:
I think we should also document what to do with outgoing requests to untrusted destinations. This is as much about egress as it is ingress.

utils/HandleUntrustedRequests.md
```diff
@@ -0,0 +1,52 @@
...
+fact that they cannot change this easily at runtime via [TraceConfig](../trace/TraceConfig.md)).
+* Use the incoming trace header (trace_id, span_id) to record a [Link][SpanDataModel] to the newly
+created `Span`.
+
+Because of the lower probability to have both incoming request and the the newly generated trace
+sampled at the same time, the library should have a way to pass the trusting information to the
+`Sampler` interface to allow users to implement smart Samplers (e.g. high sampling probability if
+the incoming untrusted request is sampled).
+
+// DO_NOT_SUBMIT: take feedback what is better 2 default samplers in TraceConfig (one for trusted
```
rakyll: User still needs to configure the entry points to tell the library incoming requests are not trusted. They can simply set a custom sampler at this point. I think we need to document it with examples for now rather than having two default samplers.

utils/HandleUntrustedRequests.md
```diff
@@ -0,0 +1,52 @@
+# Handle Untrusted Requests
+
+This document is about how to handle untrusted requests for trace and tags.
+
+## What is an untrusted request?
+An untrusted request refers to a request coming from an external untrusted source (e.g. external
+customer request).
+
+## How does OpenCensus know if a request is trusted or not?
+The OpenCensus library does not have a mechanism to determine if an incoming request is trusted
```
rakyll: Add a paragraph explaining that the starting options should take an `isTrusted bool` option.

rghetia:
> /cc @rghetia You will continue this work. Can I submit as it is and you deal with comments and make necessary changes?

sure.

bogdandrutu:
@rghetia all your now. thanks.

Issue #105, opened 2018-04-19
MonitoredResource: consider adding gcp_ and aws_ prefix to property names


songy23:
For example, we have `instance_id` in [GCE](https://github.com/census-instrumentation/opencensus-specs/blob/master/utils/MonitoredResource.md#gcp_gce_instance), and `instance_id` in [AWS EC2](https://github.com/census-instrumentation/opencensus-specs/blob/master/utils/MonitoredResource.md#aws_ec2_instance). Those two should be different properties, and I think it's more clear if we renamed them to `gcp_instance_id` and `aws_instance_id`. Same for other properties.

songy23:
This is less of an issue if we have the labels coupled with resources. Closing for now.

PR #106, opened 2018-04-19
Add prefix to all resource properties.

songy23:
Closes https://github.com/census-instrumentation/opencensus-specs/issues/105.

This makes the property names more clear. However, now the property names don't match with the Stackdriver definitions (https://cloud.google.com/monitoring/api/resources#tag_aws_ec2_instance).


songy23:
This is less of an issue if we have the labels coupled with resources. Closing for now.

PR #118, opened 2018-05-22
Introduce the Metrics data model.

g-easy:


stats/Export.md
```diff
@@ -28,6 +28,9 @@ parameter from the `View` and the aggregated data associated with those `columns
 
 The library SHOULD provide a means of retrieving the ViewData for any registered view in the system.
 
+ViewData is a language-specific data structure implementing the [Metrics data
```
songy23: I think our ultimate goal is to get rid of `ViewData` and replace it with `Metric`?

/cc @bogdandrutu

bogdandrutu: ViewData will be an intermediate step (but internal only). So from the public API will be removed but I think the implementation still needs this for accumulating data.

g-easy: How should this be re-worded? Just get rid of the mention of ViewData?

stats/Metrics.md
```diff
@@ -0,0 +1,19 @@
+# Metrics
+Metrics are a data model for what stats exporters take as input.
+
+Different exporters have different capabilities (e.g. which data types
+are supported) and different constraints (e.g. which characters are allowed in
+label keys). Metrics is intended to be a superset of what's possible, not a
+lowest common denominator that's supported everywhere.
+
+Because of this, Metrics puts minimal constraints on the data (e.g. which
+characters are allowed in kets), and code dealing with Metrics should avoid
```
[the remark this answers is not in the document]
g-easy: Fixed. Thanks!

stats/Export.md
sebright: I don't think we should mention ViewData in the spec, since it isn't visible or required.

g-easy: This file is Export.md, which existed before my PR. Not sure what the best way forward is - describe that ViewData currently exists but that it should be changed to something else in the future? What is the plan for OpenCensus Java exporters: essentially the same thing as ViewData but with a different name?

songy23: My 2 cents:

For the first question, I think most parts of Export.md need to be re-written. From exporters' perspective, there will be no `Aggregation`, `AggregationData`, `Measure`, `View` or `ViewData`. All they can see is the `MetricSet` defined in [metric proto](https://github.com/census-instrumentation/opencensus-proto/blob/master/opencensus/proto/stats/metrics/metrics.proto#L32).

For the second question, I would say yes, the Java exporters will pull `MetricSet` instead of `ViewData` from the library.

sebright: > This file is Export.md, which existed before my PR. Not sure what the best way forward is - describe that ViewData currently exists but that it should be changed to something else in the future? What is the plan for OpenCensus Java exporters: essentially the same thing as ViewData but with a different name?

Sorry, I didn't realize this was in the existing file.  I think that the note that you added is fine for now, and we can make more updates for the metrics API later.

stats/Metrics.md
```diff
@@ -0,0 +1,19 @@
...
+characters are allowed in keys), and code dealing with Metrics should avoid
+validation and sanitization of the Metrics data. Instead, pass the data to the
+backend, rely on the backend to perform validation, and pass back any errors
+from the backend.
+
+The Metrics data model is defined as
+[metrics.proto](https://github.com/census-instrumentation/opencensus-proto/blob/master/opencensus/proto/stats/metrics/metrics.proto),
+but the proto is just to illustrate the concepts. OpenCensus implementations
+don't have to use the actual proto, and can instead use a language-specific
+in-memory data structure (like `ViewData`) that captures what exporters need.
```
sebright: > language-specific
in-memory data structure (like `ViewData`)

I think that the in-memory data structures that are part of the API should still be required to use the names and fields from the Protocol Buffer, for consistency in the API across languages.  For example, I think opencensus-java will have a `Metric` class with a similar API to the class that would have been generated from the Protocol Buffer.

g-easy:
Friendly ping. What's left to do here?

sebright:
My only remaining comment was the one about the sentence mentioning ViewData in Metrics.md.

stats/Metrics.md
g-easy: I've added wording to this effect. Please take a look.

PR #121, opened 2018-06-05
Add specification for Summary Span.

rghetia:


utils/SummarySpan.md
```diff
@@ -0,0 +1,164 @@
...
+### Encoding
+
+The measurement data will be returned to the client in response trailer. The
+encoding of the same is defined for two different transport methods, Rest API
+using http and gRPC.
+
+
+#### Encoding with gRPC
+For gRPC census-server-stats-bin metadata will be sent in gRPC trailer. The
+encoding of this metadata is as per the format defined [here](https://github.com/census-instrumentation/opencensus-specs/blob/master/encodings/BinaryEncoding.md). All data is encoded
```
bogdandrutu: Shouldn't this be the link to your md and not the binary encoding?

encodings/CensusServerStatsEncoding.md
```diff
@@ -0,0 +1,45 @@
...
+
+#### Fields added in Census Server Stats version 0
+
+##### LB-Latency-Ns
+
+* optional
+* `field_id` = 0
+* `len` = 8
+
+Request processing latency observed on Load Balance. The unit is nanoseconds.
```
bogdandrutu: Mention that is an int64 little endian.

utils/SummarySpan.md
```diff
@@ -0,0 +1,164 @@
+# Objective
+A request for a typical Service goes through a cluster of Load Balancers
+before it is processed by the server. Sometimes there may be more stages in
+between. Each stage introduces some processing delay. There is also a network
+delay between these stages. These delays contribute to overall response time of
+the request. 
+
+
+From a perspective of a typical Service customer, a response time of a
+query is the time elapsed between the first Load Balancers (LB) of the provider
```
bogdandrutu: Mention that we refer to L7 LBs here, for L4 LBs we are not going to do too much now. Also for client side LB is different as well.

utils/SummarySpan.md
```diff
@@ -0,0 +1,164 @@
...
+| HIGH | MISSING | LOW | The issue is not on the server but one cannot conclusively determine if the issue is on LB, or in the network between the LB and the server, or In the network between the Client and the LB |
+| HIGH | MISSING | HIGH | The issue is on the server. Additionally, there could be an issue in other segments. |
+| HIGH | HIGH | MISSING | The issue is on LB or beyond but cannot conclusively determine if it is on the server or not. |
+
+
+### How to use Sampling bit?
+Sampling-bit is used to compare traces and measurements end-to-end. Customers
+can use the sampling-bit information to log traces and measurements on its side.
+The traces and measurement from both sides can then be used to isolate the
+problem.
```
bogdandrutu: Also mention that it is useful if the Server owner is different than the client owner because then this information helps client to find traces that are sampled on both sides (usually in this case client does not have access to the server traces).

encodings/CensusServerStatsEncoding.md
rghetia: done.

utils/SummarySpan.md
rghetia: Added a note.

utils/SummarySpan.md
rghetia: updated.

PR #122, opened 2018-06-11
Add Exemplars specification.

songy23:


stats/Export.md
```diff
@@ -15,7 +15,8 @@ The library SHOULD provide support for multiple types of Aggregations:
 type.
 * `LastValueDataDouble` and `LastValueDataInt64`: data generated for a `LastValue` aggregation based 
 on the `Measure` type.
-* `DistributionData`: data generated for a `Distribution` aggregation.
+* `DistributionData`: data generated for a `Distribution` aggregation. If there is a histogram in 
+the `Distribution`, `DistributionData` MAY contain a list of `Exemplar`s.
```
ramonza: I thought we were moving away from the "specalese" capitalization of random words =)

songy23: I see, updated.

stats/Record.md
```diff
@@ -67,6 +69,7 @@ private static final MeasureLong RPC_BYTES_SENT =
 MeasurementMap measurementMap = new MeasurementMap();
 measurementMap.put(RPC_LATENCY, 10.3);
 measurementMap.put(RPC_BYTES_SENT, 124);
+measurementMap.withSpanContext(spanContext);
```
sebright: Should the Java implementation support using the current span?  In that case, I think the no-arg `record` method should  add the spanContext implicitly.

songy23: Do you mean adding another no-arg `recordWithCurrentSpanContext()` method? I think for the existing no-arg `record()` won't interact with spanContext, to be consistent with previous behavior?

ramonza: What is the harm in sending exemplars by default? Why do we need to remain consistent with the previous behavior?

If we do, I would recommend instead that we configure exemplar support on the exporter (those that support it, which is just Stackdriver for now). 

I feel quite strongly that we shouldn't require people to change their instrumentation to get exemplars if we can avoid it.

songy23: I don't object to make it default to record again current span context, though I'm wondering will there be cases that user would only want histograms, but not linked to trace spans? If so, I think we should at least provide a method that records stats against no span context.

songy23:
PTAL. Per our meeting today, we would like to have a more generic record() method that takes a `Map<String, String>` as the attachment of `Exemplar` to avoid having dependency on Tracing APIs. The client code should take care of encoding their exemplar attachments (like TraceId, SpanId) to plain string.

stats/Record.md
```diff
@@ -69,7 +69,15 @@ private static final MeasureLong RPC_BYTES_SENT =
...
 measurementMap.record();  // reads context from thread-local.
+
+// Another example on recording against sampled SpanContext.
+SpanContext spanContext = tracer.getCurrentSpan().getContext();
+if (spanContext.getTraceOptions().isSampled()) {
+  Map<String, String> attachments = new HashMap<>();
+  // Client code needs to take care of encoding.
+  attachments.put("TraceId", encode(spanContext.getTraceId));
+  attachments.put("SpanId", encode(spanContext.getSpanId));
+  measurementMap.record(attachments);
```
sebright: I'm not sure if this makes sense, but another way to handle explicit propagation is to pass the whole gRPC context to `record`.  Then the two `record` methods would behave more similarly.  They would both allow the OpenCensus implementation to control how the exemplars are added, rather than pushing that logic into the client code.  I'm not sure how that would work in all languages, though, since opencensus-specs doesn't require a context, like the gRPC context or Go context, that references the SpanContext and TagContext and can be passed around.

stats/Record.md
```diff
@@ -68,5 +70,14 @@ MeasurementMap measurementMap = new MeasurementMap();
...
 measurementMap.put(RPC_BYTES_SENT, 124);
 measurementMap.record();  // reads context from thread-local.
+
+// Another example on recording against sampled SpanContext.
+SpanContext spanContext = tracer.getCurrentSpan().getContext();
+if (spanContext.getTraceOptions().isSampled()) {
+  Map<String, String> attachments = new HashMap<>();
+  // Client code needs to take care of encoding.
+  attachments.put("TraceId", encode(spanContext.getTraceId));
+  attachments.put("SpanId", encode(spanContext.getSpanId));
```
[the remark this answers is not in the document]
songy23: I think that's an implementation detail of the `encode()` method. Client code should decide whether or not they want the lowercase base16 encoding of TraceId/SpanId (there may also be clients that need some other encoding format).

stats/Record.md
```diff
@@ -68,5 +70,14 @@ MeasurementMap measurementMap = new MeasurementMap();
...
+  measurementMap.record(attachments);
```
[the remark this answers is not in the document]
songy23: Yes, done.

stats/Record.md
songy23: There could be other exemplar attachments that are not from the current context, like [plain string literals](https://cloud.google.com/monitoring/api/ref_v3/rest/v3/TimeSeries#exemplar). I think our goal here is to make the record() method as generic as possible, and restricting it to take only the current context seems to be against that goal.

stats/Record.md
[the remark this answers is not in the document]
ramonza: There needs to be agreement between the gRPC code and the Stackdriver exporter. Maybe we should make it a `Map<String, Object>` instead and pass the whole `spanContext`?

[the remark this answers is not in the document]
songy23: This makes sense to me, by doing that client code doesn't need to worry about encoding too. Should we change the type of [exemplar.attachment](https://github.com/census-instrumentation/opencensus-proto/pull/62/files#diff-cbf22e3bb90f48f62fe55b894a9ee40cR236) to map<string, Any> to match this?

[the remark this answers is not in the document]
ramonza: Sorry to go in circles on this but we added the Map because we didn't want an explicit dependency from the stats API to the trace API. Maybe we should have instead just made it `Object`/`interface{}` and let the stats API just pass it through? To put it another way, what purpose is the key in the map serving?

That way only gRPC and the exporters need to know about the actual type. The stats API can just treat it as an opaque pointer/object and pass it along. Then `exemplar.attachment` in the proto can go back to being just `Any`.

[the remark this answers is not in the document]
songy23: > To put it another way, what purpose is the key in the map serving?

FWIW we want a map here because there may be different types of Exemplar attachments for a single histogram bucket. For example, we can have both a set of dropped labels and a SpanContext associated with a histogram bucket. In that case, map keys are needed to differentiate attachments of different types (like a JSON object).

> That way only gRPC and the exporters need to know about the actual type. The stats API can just treat it as an opaque pointer/object and pass it along.

I think we can achieve the same goal with Map<String, Object>. The stats API doesn't need to know the actual type of the map value and can just pass it along.

stats/Record.md
```diff
@@ -44,6 +44,8 @@ measure name instead of the `Measure`.
 
 Implementations MAY define a `MeasurementMap` which describes a set of data points to be collected
 for a set of Measures. Adding this functionality may improve the efficiency of the record usage API.
+Besides, when recording Measurements, `MeasurementMap` should optionally take a map of string-object
+pairs as the attachment of exemplar.
```
dinooliva: This is more generic than originally discussed (i.e. Map<String, String>) but I think makes sense. Let's discuss it in the cross-language sync in case non-gc'd languages may have concerns about object lifetimes.

stats/Record.md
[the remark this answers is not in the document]
sebright: I think Object is too generic, because exporters won't be able to pass it along without knowing what it represents, since it's not possible to serialize unknown Objects.  Using String values gives exporters the flexibility to either look for specific keys that they understand, or export the whole map.

stats/Record.md
sebright: Then it seems like we should try to make the implicit propagation version of `record` more generic, too, though I'm not sure how that would work.  Maybe both `record` methods should take a `Map<String, String>`, and the implicit propagation version could merge the input map with the keys-value pairs representing the current span.

stats/Record.md
songy23: Per our discussion today I've reverted the change to use Map<String, String> instead.

stats/Record.md
```diff
@@ -44,11 +44,15 @@ measure name instead of the `Measure`.
 
 Implementations MAY define a `MeasurementMap` which describes a set of data points to be collected
 for a set of Measures. Adding this functionality may improve the efficiency of the record usage API.
+Besides, when recording Measurements, `MeasurementMap` should optionally take a map of string
```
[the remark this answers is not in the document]

stats/Record.md
```diff
@@ -44,11 +44,15 @@ measure name instead of the `Measure`.
 
 Implementations MAY define a `MeasurementMap` which describes a set of data points to be collected
 for a set of Measures. Adding this functionality may improve the efficiency of the record usage API.
+Besides, when recording Measurements, `MeasurementMap` should optionally take a map of string 
+key-value pairs as the attachment of exemplar.
```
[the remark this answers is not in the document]
songy23: I think we defined `Exemplar` as `value` + `timestamp` + `attachments` in https://github.com/census-instrumentation/opencensus-proto/pull/62. `value` is included in `measurement`, `timestamp` is added by the library, so we need to pass in the `attachments`, which is the string map.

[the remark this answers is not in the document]
songy23: I see, would it be better to rephrase this to "MeasurementMap should optionally take a map of string key-value pairs to record exemplar"?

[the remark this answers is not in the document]
dinooliva: That works for me - nit: "to record an exemplar" or "to record exemplars"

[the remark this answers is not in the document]
songy23: Yep, updated.

stats/Record.md
```diff
@@ -44,11 +44,15 @@ measure name instead of the `Measure`.
...
 for a set of Measures. Adding this functionality may improve the efficiency of the record usage API.
+Additionally, when recording Measurements, `MeasurementMap` should optionally take a map of string 
+key-value pairs to record an exemplar.
 
 ## Recording Stats
 
 Users should record Measurements against a context, either an explicit context or the implicit 
 current context. Tags from the context are recorded with the Measurements if they are any.
+Note that there is no implicit recording for exemplars. If you want to record a `Measurement`
+against an exemplar, you have to explicitly pass the exemplar attachments (a string-string map).
```
dinooliva: You should word this similarly (i.e. getting rid of the term attachments).

stats/Record.md
```diff
@@ -68,5 +72,14 @@ MeasurementMap measurementMap = new MeasurementMap();
...
 measurementMap.record();  // reads context from thread-local.
+
+// Another example on recording against sampled SpanContext.
+SpanContext spanContext = tracer.getCurrentSpan().getContext();
+if (spanContext.getTraceOptions().isSampled()) {
+  Map<String, String> map = new HashMap<>();
+  // Client code needs to take care of encoding.
+  map.put("TraceId", encode(spanContext.getTraceId()));
+  map.put("SpanId", encode(spanContext.getSpanId()));
+  measurementMap.record(tagContext, map);
```
adriancole: can someone mention how an existing metrics system might handle this data as it is inherently unlimited cardinality. For example, is there an exemplar feature somewhere? I get where this is going, but bottom-up I'm kindof guessing.

bogdandrutu: @adriancole here are some examples of data we want to record:
1) TraceId/SpanId this will allow the backend to display on their UI information about traces within a percentile (usually) (e.g. "Here are 99%tile traces").
2) Dropped tags. Let's assume that in the tag map/set/context we have 20 unique tags and we break down stats only by 4 of them. We record the rest of 16 as dropped tags. This can be used as example of requests tags for different percentiles.

The first example is what we call correlation between metrics and traces.

PR #129, opened 2018-06-28
Stats: Update the Java code snippet on recording Exemplar.

songy23:
The idea is essentially the same. The builder-style API (`putAttachment()`) should make the API easier to use, compared to overloading and having multiple versions of `record()`.

This matches the current API design in Java (https://github.com/census-instrumentation/opencensus-java/pull/1285).

### DIFF
--- a/stats/Record.md
+++ b/stats/Record.md
@@ -76,10 +76,10 @@ measurementMap.record();  // reads context from thread-local.
 // Another example on recording against sampled SpanContext.
 SpanContext spanContext = tracer.getCurrentSpan().getContext();
 if (spanContext.getTraceOptions().isSampled()) {
-  Map<String, String> map = new HashMap<>();
   // Client code needs to take care of encoding.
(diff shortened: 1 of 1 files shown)


songy23:
> I think we should also describe "attachments" where we describe exemplars in the spec.

Make sense, I added some explanation in the section above.

stats/Record.md
```diff
@@ -45,7 +45,8 @@ measure name instead of the `Measure`.
 Implementations MAY define a `MeasurementMap` which describes a set of data points to be collected
 for a set of Measures. Adding this functionality may improve the efficiency of the record usage API.
 Additionally, when recording Measurements, `MeasurementMap` should optionally take a map of string 
-key-value pairs to record an exemplar.
+key-value pairs to record an exemplar. The string map is called attachments and represents the
```
dinooliva: Consider quoting 'attachments'

PR #136, opened 2018-07-12
Allow customizing gRPC propagation

ramonza:
This change is to support environments where gRPC traffic is treated as HTTP/2. Concretely to support propagating B3 tracing headers to support tracing in Istio/Envoy. I could also imagine it being useful also to gradually transition to TraceContext for gRPC (if this is something that we want to do).

### DIFF
--- a/trace/gRPC.md
+++ b/trace/gRPC.md
@@ -4,7 +4,7 @@ This document explains tracing of gRPC requests with OpenCensus.
 
 ## Spans
 
-Implementations MUST create a span, when the gRPC call starts, for the client and a span for the 
+Implementations MUST create a span, when the gRPC call starts, for the client and a span for the
 server.
 
 Span name is formatted as (also known as full gRPC method name):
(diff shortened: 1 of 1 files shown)


ramonza:
See: https://github.com/census-instrumentation/opencensus-go/pull/826

rakyll:
Previously we had one user who needed B3 propagation for gRPC for legacy reasons (they were using Envoy) and we haven't heard from any other user that B3 is critical for them.

I asked @bogdandrutu a while ago to comprehensively document the grpc-trace-bin and I am happy to do it if he doesn't want to. If the binary format is the new canonical format, we can ask Envoy to implement it or support both B3 and binary headers for the migration period.

Asking users to set one more thing correctly might not be ideal if we are trying to unify the propagation formats.

Note: This PR requires brief explanation why this change is required.

adriancole:
FYI B3 has been used in zipkin related tracing since gRPC 1.3, not just java but I think also go and javascript. We made a change to the java library to allow it to interop with stuff natively, but a similar change would need to occur in the various libraries that use B3 for gRPC (as well a rollout of such change)

https://github.com/openzipkin/brave/tree/master/instrumentation/grpc#grpc-propagation-format-census-interop

basvanbeek:
I will add support to zipkin-go for `grpc-trace-bin` as it currently only supports lowercased B3

ramonza:
@basvanbeek adding support for grpc-trace-bin to zipkin-go (which is not associated with OpenCensus) isn't what's being asked for here and doesn't really help people who have infrastructure that only understands B3 headers.

basvanbeek:
It's not asked but it does help... There is only a few Zipkin libraries which have been supporting lowercased B3 for gRPC, one is zipkin-go. I'd like to adhere to the more standard gRPC header as provided by OpenCensus for gRPC propagation interop. For most this is a greenfield situation and moving toward a standard for gRPC is very beneficial. As you can see @adriancole also mentioned having added support for it in brave.

ramonza:
Alright, just didn't see the relevance for this issue or OpenCensus but I don't have any objections to that, or really any opinion at all.

Just an aside: all HTTP headers are case-insensitive so lower-case B3 should be the same as B3 or library parsing the headers is not HTTP spec compliant.

basvanbeek:
I know HTTP headers are case-insensitive. I specifically call it lower-case B3 for gRPC as we've noticed people messing up parsing of raw headers in the past and we've been communicating all lowercase for gRPC so people won't do strange things.

Spec is one... implementation is another... hence me being very explicit there

ramonza:
Closing this for now, will re-open if/when it becomes a priority again.

terinjokes:
@rakyll @ramonza Being that (prospective) user circling back on this, I still care. :smile:

PR #137, opened 2018-07-13
Add spec for YAML view definition

ramonza:
I am proposing this to be used both in the libraries for now and eventually for OpenCensus service to support runtime view definition.

### DIFF
--- a/stats/ViewConfig.md
+++ b/stats/ViewConfig.md
@@ -3,6 +3,9 @@
 This document describes the YAML configuration for views.
 
 Implementations should provide a way to load this configuration to define views.
+Implementations may provide facilities to:
+* watch the configuration file for changes and reload the views
+* reload the file explicitly by function call
 
 ## Schema
 
--- a/stats/ViewConfig.md
+++ b/stats/ViewConfig.md
@@ -49,10 +49,38 @@ Each view definition contains the following fields:
 
 ### Aggregation object
 
-| Field        | Type             | Required | Comments                                                                                |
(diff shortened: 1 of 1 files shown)


bogdandrutu:
The yaml definition looks great, I have a bit of troubles to understand the "runtime support" part. 

Can you please explain me how that will work, I have in mind some use-cases:
1) When a new process start - we need to load a set of views (we can use a yaml file to configure that).
2) Add/Remove one view at runtime - How is this going to be supported in this design? Do I have to get all defined views, then modify what I need and send the whole list of views again?

ramonza:
@bogdandrutu This proposal assumes this will be read from a configuration file. I added a section that says implementations may support reloading the file at runtime.

So yes, if you want to add a view you change the file and then reload it. Concretely, in most setups you will probably store this file somewhere and it will be deployed to the node. For example you could mount it as a configmap in K8s.

stats/ViewConfig.md
```diff
@@ -0,0 +1,126 @@
+# View Definition
+
+This document describes the YAML configuration for views.
+
+Implementations should provide a way to load this configuration to define views.
+Implementations may provide facilities to:
+* watch the configuration file for changes and reload the views
```
rakyll: Is the file here any resource? E.g. metadata from a metadata service?

I'd not call it file to be more precise.

ramonza:
I am going to rationalize this with the view proto. Stay tuned.

ramonza:
Closing this for now, I think it should be handled by making the protobuf suitable for use as config or defining new protobuf that are useful for config.

Issue #138, opened 2018-07-13
Buffer limit API for message events, links, and annotations?


jgracenin:
I have a Go service with a long running stream that had a memory leak due to unbounded storage of 
message events. This was a known limitation: census-instrumentation/opencensus-go#401 

gRPC's MAX_CONNECTION_AGE would help, but I also made a quick fork with a limit on message events, annotations, and links: census-instrumentation/opencensus-go#824 .

Do any other language implementations already limit these internal structures yet (events, links annotations) yet?

The PR also sparked a larger discussion on more granular limits for some future features (e.g. classified events, links annotations), but that's a larger API change and is probably better to be discussed here.

songy23:
> Do any other language implementations already limit these internal structures yet (events, links annotations) yet?

Yes, in Java we have some [default limit for them](https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/trace/config/TraceParams.java#L40-L43):
```java
  private static final int DEFAULT_SPAN_MAX_NUM_ATTRIBUTES = 32;
  private static final int DEFAULT_SPAN_MAX_NUM_ANNOTATIONS = 32;
  private static final int DEFAULT_SPAN_MAX_NUM_MESSAGE_EVENTS = 128;
  private static final int DEFAULT_SPAN_MAX_NUM_LINKS = 128;
```

Users can override these numbers with [TraceConfig.updateActiveTraceParams](https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/trace/config/TraceConfig.java#L42).

rakyll:
@songy23, can we update https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/TraceConfig.md to represent this? So, other languages can refer to the spec when implementing.

songy23:
Sure, working on a PR now.

g-easy:
https://github.com/census-instrumentation/opencensus-specs/issues/138#issuecomment-404962790 FYI C++ has the same.

PR #139, opened 2018-07-13
Tracing: Add specifications on default TraceParams.

songy23:
Related to https://github.com/census-instrumentation/opencensus-specs/issues/138.

/cc @bogdandrutu


songy23:
> 128 links sounds very high. what is a use case for tens of links much less over a hundred in the same span?

@bogdandrutu Could you answer this question?

bogdandrutu:
The size of links is small so no memory issue to support 128, but probably the number is too high for real world use-cases.

@songy23 I would reduce that to 32 probably. The reason for this is that if you have some kind of batching you likely have >10 batched operations (not always but sometime).

songy23:
SG, updated the limit on `Link`s to 32. Will update Java default params as well.

adriancole:
thx folks

Issue #140, opened 2018-07-16
Providing tags without context objects


fabxc:
Contexts are currently the only way to provide tag values (AFAICT), which makes sense for the majority of metrics, which are scoped to a some kind of request and certainly for traces.

For metrics there are many cases though, where one just wants to measure some state of application internals, e.g. the length of a queue.
It would be helpful if one could initialize a static set of tags somehow (e.g. queue identifier) without forcing a context object into all APIs for the sole purpose of setting a metric tag. It may actually force developers to significantly refactor and break their APIs if they want to add OC instrumentation to existing code.

I think similar issues would arise in practice for an OC logging API.

@jkohen

ramonza:
As an example, do you mean (in Go) something hypothetically like:

```go

stats.Record(ctx, myMeasure.M(1), stats.WithTag(TagKeyQueueName, "my-queue-1"))
```

songy23:
In Java you can use [MeasureMap.record(TagContext tags)](https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/stats/MeasureMap.java#L91), which takes additional tags when recording without putting them into context.

fabxc:
Looks like Java supports what I was thinking of.
I was thinking of a similar Go API but currently the measure argument is already variadic, so that would mean a breaking API change. Though I think recording multiple measure at once is infrequent enough for this to be ok.

fabxc:
I see, that makes sense. Probably best to just add a separate methods that takes an additional tag slice then?

rakyll:
I also think adding a separate function would be the best. It would be also easier to document the difference. What about the following?

```
func RecordWithTags(ctx context.Context, tags map[tag.Key]string, m ....Measurement)
```

ramonza:
This is an opencensus-go specific discussion, let's continue there.

songy23:
I think we already have that specification?

> [Users should record Measurements against a context, either an explicit context or the implicit current context.](https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/Record.md#recording-stats)

Maybe we can add another sentence says users can put additional tags in the explicit context.

rakyll:
Maybe we can say something like "It should be easy to record with non-propagated tags." and it is implementation detail how you achieve it.

songy23:
> Maybe we can say something like "It should be easy to record with non-propagated tags." and it is implementation detail how you achieve it.

[...]

PR #145, opened 2018-07-17
Add a note about recording with extra tags.

songy23:
Closes https://github.com/census-instrumentation/opencensus-specs/issues/140.

### DIFF
--- a/stats/Record.md
+++ b/stats/Record.md
@@ -77,6 +77,12 @@ measurementMap.put(RPC_LATENCY, 10.3);
 measurementMap.put(RPC_BYTES_SENT, 124);
 measurementMap.record();  // reads context from thread-local.
 
+MeasurementMap measurementMap = new MeasurementMap();
+measurementMap.put(RPC_LATENCY, 15);
+measurementMap.put(RPC_BYTES_SENT, 200);
+measurementMap.record(
+    Tags.getTagger().currentBuilder().put(myKey, myValue).build());  // record against an extra tag.
+
 // Another example on recording against sampled SpanContext.
 SpanContext spanContext = tracer.getCurrentSpan().getContext();
 if (spanContext.getTraceOptions().isSampled()) {
--- a/stats/Record.md
+++ b/stats/Record.md
@@ -51,7 +51,10 @@ contextual information of an exemplar, for example trace id, span id or dropped 
 ## Recording Stats
 
 Users should record Measurements against a context, either an explicit context or the implicit 
-current context. Tags from the context are recorded with the Measurements if they are any.
+current context. Tags from the context are recorded with the Measurements if they are any. 
+When recording against an explicit context, implmentations should allow users to add extra tags,
+and those tags should be non-propagating.
+
 Note that there is no implicit recording for exemplars. If you want to record a `Measurement`
 against an exemplar, you have to explicitly pass a string-string map.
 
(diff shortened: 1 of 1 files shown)


stats/Record.md
```diff
@@ -51,7 +51,10 @@ contextual information of an exemplar, for example trace id, span id or dropped
 ## Recording Stats
 
 Users should record Measurements against a context, either an explicit context or the implicit 
-current context. Tags from the context are recorded with the Measurements if they are any.
+current context. Tags from the context are recorded with the Measurements if they are any.
```
bogdandrutu: Please change the example to show this as well.

stats/Record.md
```diff
@@ -74,6 +77,12 @@ measurementMap.put(RPC_LATENCY, 10.3);
 measurementMap.put(RPC_BYTES_SENT, 124);
 measurementMap.record();  // reads context from thread-local.
 
+MeasurementMap measurementMap = new MeasurementMap();
+measurementMap.put(RPC_LATENCY, 15);
+measurementMap.put(RPC_BYTES_SENT, 200);
+measurementMap.record(
+    Tags.getTagger().currentBuilder().put(myKey, myValue).build());  // record against an extra tag.
```
sebright: It might be clearer to split the code into four blocks, shared measure constants and three examples of recording stats.  Then some of the code comments could be moved above the blocks of code that they describe.

songy23: Sounds good, updated.

sebright: I meant that the code comment at the top could be moved outside of the code block, as a description for the whole code block, though the current formatting is also fine.

stats/Record.md
```diff
@@ -64,16 +67,33 @@ argument. e.g. `record(List<Measurement>)` or `record(...Measurement)`.
...
+```
 
+```java
+// Record against explicit context.
+MeasurementMap measurementMap = new MeasurementMap();
+measurementMap.put(RPC_LATENCY, 15);
+measurementMap.put(RPC_BYTES_SENT, 200);
+TagValue value = TagValue.create("some value");
+measurementMap.record(
+    Tags.getTagger().currentBuilder().put(MY_KEY, value).build());  // record against an extra tag.
```
bogdandrutu: I think this is exactly the same problem as Go. what we want in that is something like:
private static final Map<TagKey, TagValue> extraTags = ...
measurementMap.recordWithTags(currentTagContext, extraTags)

sebright: What is the benefit of the API that takes a Map?  It looks like it would have the same behavior as the example in this PR, except that it doesn't allow the library to skip creating the map as an optimization when OpenCensus isn't enabled.

bogdandrutu: Please comment on the main issue.

songy23:
Going to merge this PR if there are no comments/concerns by EOD.

stats/Record.md
```diff
@@ -51,7 +51,10 @@ contextual information of an exemplar, for example trace id, span id or dropped
 ## Recording Stats
 
 Users should record Measurements against a context, either an explicit context or the implicit 
-current context. Tags from the context are recorded with the Measurements if they are any.
+current context. Tags from the context are recorded with the Measurements if they are any. 
+When recording against an explicit context, implmentations should allow users to add extra tags,
+and those tags should be non-propagating.
```
ramonza: We haven't really defined what non-propagating means. I think it would be enough to say here that the extra tags are not added to the context.

songy23: Sounds good, updated.

PR #158, opened 2018-08-10
Add details about supporting qps based sampler.

bogdandrutu:
This idea belongs to a different Google engineer who does not have a github account to link.

### DIFF
--- a/trace/Sampling.md
+++ b/trace/Sampling.md
@@ -14,9 +14,9 @@ The Sampling bit is always set only at the start of a Span, using a `Sampler`
 * `Probability` - sampler that tries to uniformly sample traces with a given probability. When 
 applied to a child `Span` of a **sampled** parent `Span`, the child `Span` keeps the sampling
 decision.
-* `QPS` - sampler that tries to sample traces based on QPS (e.g. 1 trace/second). When applied to a
-child `Span` of a **sampled** parent `Span`, the child `Span` keeps the sampling decision. For
-implementation details see [this](#qps-sampler-implementation-details)
+* `RateLimiting` - sampler that tries to sample with a rate per time window (0.1 traces/second). 
+When applied to a child `Span` of a **sampled** parent `Span`, the child `Span` keeps the sampling 
+decision. For implementation details see [this](#ratelimiting-sampler-implementation-details)
 
 ### How can users control the Sampler that is used for sampling?
 There are 2 ways to control the `Sampler` used when the library samples:
@@ -38,7 +38,7 @@ The OpenCensus library samples based on the following rules:
    * If a "span-scoped" `Sampler` is provided, use it to determine the sampling decision.
    * Else keep the sampling decision from the parent.
 
-### QPS sampler implementation details
+### RateLimiting sampler implementation details
 The problem we are trying to solve is:
 1. Getting QPS based sampling.
 2. Providing real sampling probabilities.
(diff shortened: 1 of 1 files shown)


trace/Sampling.md
```diff
@@ -12,7 +12,11 @@ The Sampling bit is always set only at the start of a Span, using a `Sampler`
 * `AlwaysSample` - sampler that makes a "yes" decision every time.
 * `NeverSample` - sampler that makes a "no" decision every time.
 * `Probability` - sampler that tries to uniformly sample traces with a given probability. When 
-applied to a child `Span` of a **sampled** parent `Span`, the child `Span` keeps the sampling decision.
+applied to a child `Span` of a **sampled** parent `Span`, the child `Span` keeps the sampling
+decision.
+* `QPS` - sampler that tries to sample traces based on QPS (e.g. 1 trace/second). When applied to a
```
ramonza: I'm not sure `QPS` is a good name for this. It's common in Google to use QPS to mean rate of anything (not just queries) but I think this is less common elsewhere in my experience, but I would be interested to hear non-Google perspectives on this. What about "fixed rate sampler"?

trace/Sampling.md
```diff
@@ -33,3 +37,26 @@ The OpenCensus library samples based on the following rules:
 3. If the span is a child of a local `Span` the sampling decision will be:
    * If a "span-scoped" `Sampler` is provided, use it to determine the sampling decision.
    * Else keep the sampling decision from the parent.
+
+### QPS sampler implementation details
+The problem we are trying to solve is:
+1. Getting QPS based sampling.
+2. Providing real sampling probabilities.
```
ramonza: What does this mean? Do you mean that we can attach to each span the actual sampling probability in use?

trace/Sampling.md
yurishkuro: In Jaeger we call them rate limiting samplers

ramonza: +1 for rate limiting - my previous suggestion of fixed rate isn't good because if the application produces fewer than this number of spans, the rate will actually be lower.

adriancole:
I think Amazon calls it reservoir sampler

ramonza:
Reservoir sampling is an algorithm to obtain a fixed size sample from a continuous stream where each item (usually) has a fixed chance of being included in the sample. Seems like it could be useful as a technique for implementing this but I don't like it as a name since it is more descriptive of the "how" rather than the "what".

trace/Sampling.md
ramonza: @nevinheintze mentioned offline that we might want to set the default rate significantly lower than 1 trace/sec to avoid issues when there are a very large number of tasks - maybe 0.1 trace/sec.

trace/Sampling.md
bogdandrutu: That was my idea that we may want to do this.

trace/Sampling.md
bogdandrutu: Agreed.

trace/Sampling.md
```diff
@@ -33,3 +37,26 @@ The OpenCensus library samples based on the following rules:
 3. If the span is a child of a local `Span` the sampling decision will be:
    * If a "span-scoped" `Sampler` is provided, use it to determine the sampling decision.
    * Else keep the sampling decision from the parent.
+
+### QPS sampler implementation details
```
ramonza: An alternative here would be to use reservoir sampling I think, should we explore that? Sounds like it's what is used to accomplish this for other systems.

bogdandrutu: Sure, I would look into that. The advantage of this algorithm is that we update only one global atomic variable, happy to see a small pseudocode for the reservoir.

adriancole:
FWIW rate limited is less jargon anyway so good by me

PR #161, opened 2018-08-22
Add a new tag http_server_handler

ramonza:
We need a low cardinality tag that we can use to break down views.
This might not always be available, but when it is it is very
useful.


ramonza:
Open to other suggestions about the name of this tag. Also, although not included here I think we should consider changing the default server views to include this tag. The default views are currently not implemented in Go -- no sure if they are implemented yet in any other language?

stats/HTTP.md
```diff
@@ -76,22 +76,27 @@ Server stats are recorded for each individual HTTP request, including for each r
...
-| http_server_path   | URL path (not including query string)                               |
-| http_server_status | HTTP server status code returned, as an integer e.g. 200, 404, 500. |
-| http_server_host   | Value of the request Host header                                    |
+| Tag name            | Description                                                         |
+|---------------------|---------------------------------------------------------------------|
+| http_server_method  | HTTP method, capitalized (i.e. GET, POST, PUT, DELETE, etc.)        |
+| http_server_path    | URL path (not including query string)                               |
+| http_server_status  | HTTP server status code returned, as an integer e.g. 200, 404, 500. |
+| http_server_host    | Value of the request Host header                                    |
+| http_server_handler | Logical name of the handler of this request (see below)               |
```
bogdandrutu: I am not sure I understand what is the value of this tag, or how does the library set it.

ramonza: Currently we don't have a low cardinality tag besides method. This tag would be set to something low cardinality that indicates how the request was processed on the server side. I wanted it to be fairly generic, since different languages and web frameworks may have different concepts of this but it something like a servlet name in Java or in Go a URL pattern matched. Currently the HTTP views are not very useful since there is no safe (low cardinality) way to distinguish different handlers, besides based on the method. As mentioned here, the path is not necessarily low cardinality.

ramonza: Method is the HTTP method: so GET/POST/DELETE etc.

bogdandrutu: I don't disagree with having this. I would like to say how this tag will be set. User has to manually set it as an option in the plugin? We automatically set it to some values? Callback that users provide to extract parts from URL, etc.

The only thing that I want to know is how the library should implement the support for this tag.

ramonza: Here is how I propose to do it in Go:
https://github.com/census-instrumentation/opencensus-go/pull/879

bogdandrutu: Couple of suggestions:
* I think currently the definition of what this tags mens is combined with the explanation about high-cardinality maybe split the explanation of how this tag should be set in a separate phrase.
* I don't really like the name but I don't have a better suggestion. If you have more ideas can you please put them as options in the comment so we can decide which is the best option.

ramonza: I made it a separate paragraph. My only other suggestion is `http_server_route` - I have a slight preference this over `http_server_handler`.

ramonza: @rakyll what do you think of `http_server_route`?

bogdandrutu: http_server_route looks good to me and slightly better. What others think?

rakyll: http_server_route is more consistent with the trace attribute key, +1 to route.

ramonza:
We decided to rename the tag to http_server_route.

PR #162, opened 2018-08-23
Add resource specs

fabxc:
This spec aims to cover my proposal for a resource package in the core library ([Go reference implementation](https://github.com/census-instrumentation/opencensus-go/pull/878)).
I dropped the existing spec that was covering specific cloud vendor auto-detection since part of the proposal is that those live independently in `census-ecosystem` or entirely elsewhere.

I took my best guesses on MAY vs. SHOULD vs. MUST based on existing specs but don't feel strongly in most cases. Any guidance is appreciated.

### DIFF
--- a/resource/README.md
+++ b/resource/README.md
@@ -1,0 +1,10 @@
+# OpenCensus Library Resource Package
+This documentation serves to document the "look and feel" of the open source resource package.
+It describes they key types and overall behavior.
+
+The primary purpose of resources as a first-class concept in the core library is decoupling
+of discovery of resource information from exporters. This allows for independent development
+of either and easy customization for users that need to integrate with closed source environments.
+
+## Main APIs
+* [Resource](Resource.md)
--- a/resource/README.md
+++ b/resource/README.md
@@ -2,9 +2,13 @@
 This documentation serves to document the "look and feel" of the OpenCensus resource package.
 It describes their key types and overall behavior.
 
+The resource library primarily defines a type "Resource" that captures information about the
+entity for which stats or traces are recorded. For example, metrics exposed by a Kubernetes
+container can be linked to a resource that specifies the cluster, namespace, pod, and container name.
 The primary purpose of resources as a first-class concept in the core library is decoupling
 of discovery of resource information from exporters. This allows for independent development
 of easy customization for users that need to integrate with closed source environments.
+
 
 ## Main APIs
 * [Resource](Resource.md)
--- a/resource/README.md
+++ b/resource/README.md
@@ -1,10 +1,10 @@
 # OpenCensus Library Resource Package
-This documentation serves to document the "look and feel" of the open source resource package.
-It describes they key types and overall behavior.
+This documentation serves to document the "look and feel" of the OpenCensus resource package.
+It describes their key types and overall behavior.
 
 The primary purpose of resources as a first-class concept in the core library is decoupling
 of discovery of resource information from exporters. This allows for independent development
-of either and easy customization for users that need to integrate with closed source environments.
+of easy customization for users that need to integrate with closed source environments.
 
 ## Main APIs
 * [Resource](Resource.md)
--- a/resource/Resource.md
+++ b/resource/Resource.md
@@ -1,13 +1,13 @@
 # Resource API Overview
-The resource library primarily defines a type that captures about information about the entity
+The resource library primarily defines a type that captures information about the entity
 for which stats or traces are recorded. It further provides a framework for detection of
 resource information from the environment and progressive population as signals propagate
-from the core instrumentation library to the a backend's exporter.
+from the core instrumentation library to a backend's exporter.
 
 ## Resource type
 A `Resource` describes the entity for which a signal was collected through two fields:
 * `type`: a string which describes a well-known type of entity. It SHOULD be namespaced
-to avoid collisions across different environment, e.g. `kubernetes.io/container`,
+to avoid collisions across different environment, e.g. `k8s.io/container`,
 `cloud.google.com/gce/instance`. 
 * `labels`: a dictionary of labels with string keys and values that provide information
 about the entity.
@@ -17,7 +17,7 @@ single slashes. All other characters MUST be alphanumeric. Label values MAY cont
 unicode character.
 
 Implementations MAY define a `Resource` data type, constructed from the parameters above.
-Resource MAY have getters for retrieving all the information used in `Resource` definition.
+`Resource` MUST have getters for retrieving all the information used in `Resource` definition.
 
 Example in Go:
 ```go
@@ -29,7 +29,7 @@ type Resource {
 
 ## Populating resources
 Resource information MAY be populated at any point between startup of the instrumented
-application and passing it the a backend-specific exporter. This explicitly includes
+application and passing it to a backend-specific exporter. This explicitly includes
 the path through future OpenCensus components such as agents or services.
 
 For example, process-identifying information may be populated through the library while
@@ -46,6 +46,10 @@ Two environment varibales are used:
 The key MUST be seperated from the value by a single `=`. Values MAY be quoted with a single
 leading and trailing `"`. If a values contains whitespaces, `=`, or `"` characters it MUST be
 quoted and `"` characters MUST be escaped.
+
+For example:
+* `OC_RESOURCE_TYPE=k8s.io/container`
+* `OC_RESOURCE_LABELS=k8s.io/pod_name="pod-xyz-123",k8s.io/container_name="c1",k8s.io/namespace="default"`
 
 Population from environment variables MUST be the first applied detection process unless
 the user explicitly overwrites this behavior.
@@ -90,7 +94,7 @@ For example, from a resource object
 
 ```javascript
 {
-	"type": "kubernetes.io/container",
+	"type": "k8s.io/container",
 	"labels": {
 		// Populated from VM environment through auto-detection library.
 		"cloud.google.com/gce/instance_id": "instance1",
@@ -98,9 +102,9 @@ For example, from a resource object
 		"cloud.google.com/project_id": "project1",
 		"cloud.google.com/gce/attributes/cluster_name": "cluster1",
 		// Populated through OpenCensus resource environment varibales.
-		"kubernetes.io/namespace": "ns1",
-		"kubernetes.io/pod_name": "pod1",
-		"kubernetes.io/container_name": "container1",
+		"k8s.io/namespace": "ns1",
+		"k8s.io/pod_name": "pod1",
+		"k8s.io/container_name": "container1",
 	},
 }
 ```
@@ -128,6 +132,6 @@ Those would generally be based on community-supported detection integrations mai
 
 Additionally, exporters SHOULD provide configuration hooks for users to provide their own
 translation unless the exporter's backend does not support resources at all. For such backends,
-exporters SHOULD allow attaching converting resource labels to metric labels.
+exporters SHOULD allow attaching converting resource labels to metric tags.
 
 [census-ecosystem]: https://github.com/census-ecosystem
--- a/resource/Resource.md
+++ b/resource/Resource.md
@@ -1,20 +1,19 @@
 # Resource API Overview
 The resource library primarily defines a type that captures information about the entity
-for which stats or traces are recorded. It further provides a framework for detection of
+for which stats or traces are reported. It further provides a framework for detection of
 resource information from the environment and progressive population as signals propagate
 from the core instrumentation library to a backend's exporter.
 
 ## Resource type
 A `Resource` describes the entity for which a signal was collected through two fields:
-* `type`: a string which describes a well-known type of entity. It SHOULD be namespaced
-to avoid collisions across different environment, e.g. `k8s.io/container`,
-`cloud.google.com/gce/instance`. 
+* `type`: an optional string which describes a well-known type of resource.
 * `labels`: a dictionary of labels with string keys and values that provide information
 about the entity.
 
-Type strings and label keys MAY start with a domain name and MAY be further namespaced through
-single slashes. All other characters MUST be alphanumeric. Label values MAY contain any valid
-unicode character.
+Type, label keys, and label values MUST contain only printable ASCII (codes between 32
+and 126, inclusive) and not exceed 256 characters.
+Type and label keys MUST have a length greater than zero. They SHOULD start with a domain
+name and separate hierarchies with `/` characters, e.g. `k8s.io/namespace/name`.
 
 Implementations MAY define a `Resource` data type, constructed from the parameters above.
 `Resource` MUST have getters for retrieving all the information used in `Resource` definition.
@@ -42,14 +41,12 @@ that may not be detectable automatically through available integration libraries
 
 Two environment variables are used:
 * `OC_RESOURCE_TYPE`: defines the resource type. Leading and trailing whitespaces are trimmed.
-* `OC_RESOURCE_LABELS`: defines resource labels as a comma-seperated list of key/value pairs.
-The key MUST be seperated from the value by a single `=`. Values MAY be quoted with a single
-leading and trailing `"`. If a values contains whitespaces, `=`, or `"` characters it MUST be
-quoted and `"` characters MUST be escaped.
+* `OC_RESOURCE_LABELS`: defines resource labels as a comma-seperated list of key/value pairs
+(`[ <key>="value" [ ,<key>="<value>" ... ] ]`). `"` characters in values MUST be escaped with `\`.
 
 For example:
 * `OC_RESOURCE_TYPE=k8s.io/container`
-* `OC_RESOURCE_LABELS=k8s.io/pod_name="pod-xyz-123",k8s.io/container_name="c1",k8s.io/namespace="default"`
+* `OC_RESOURCE_LABELS=k8s.io/pod/name="pod-xyz-123",k8s.io/container/name="c1",k8s.io/namespace/name="default"`
 
 Population from environment variables MUST be the first applied detection process unless
 the user explicitly overwrites this behavior.
@@ -84,13 +81,12 @@ func ChainedDetector(...Detector) Detector
 ```
 
 ### Updates
-Over the runtime of an application, resource information may change in some cases. To
-allow exporters to incorporate those changes, they SHOULD take a resource getter as a
-configuration parameter. The exporter SHOULD periodically rerun the getter and incorporate
-any changes into the data it exporters.
-If the getter returns an error, the exporter SHOULD log the error and proceed with the
-last known resource information. It SHOULD run the getter once on initialization and
-fail if it returns an error.
+Over the runtime of an application, attributes of a resource may change in some cases.
+OpenCensus's resource representation is focused on providing static, uniquely identifying
+information and thus those mutable attributes SHOULD NOT be included in the resource
+representation.
+Resource type and labels MUST NOT be mutated after initialization. Any changes MUST be
+effectively be treated as a different resource and any associated signal state MUST be reset.
 
 ## Exporter translation
 A resource object MUST NOT be mutated further once it is passed to a backend-specific exporter.
@@ -112,9 +108,9 @@ For example, from a resource object
 		"cloud.google.com/project_id": "project1",
 		"cloud.google.com/gce/attributes/cluster_name": "cluster1",
 		// Populated through OpenCensus resource environment variables.
-		"k8s.io/namespace": "ns1",
-		"k8s.io/pod_name": "pod1",
-		"k8s.io/container_name": "container1",
+		"k8s.io/namespace/name": "ns1",
+		"k8s.io/pod/name": "pod1",
+		"k8s.io/container/name": "container1",
 	},
 }
 ```
@@ -137,7 +133,7 @@ resource type with well-known identifiers specific to its API:
 ```
 
 For another, hyopthetical, backend a simple unique identifier might be constructed instead
-by its expoerter:
+by its exporter:
 
 ```
 cluster1/ns1/pod1/container1
--- a/resource/Resource.md
+++ b/resource/Resource.md
@@ -36,11 +36,11 @@ For example, process-identifying information may be populated through the librar
 an agent attaches further labels about the underlying VM, the cluster, or geo-location.
 
 ### From environment variables
-Population of resource information from environment varibales MUST be provided by the
+Population of resource information from environment variables MUST be provided by the
 core library. It provides the user with an ubiquitious way to manually provide information
 that may not be detectable automatically through available integration libraries.
 
-Two environment varibales are used:
+Two environment variables are used:
 * `OC_RESOURCE_TYPE`: defines the resource type. Leading and trailing whitespaces are trimmed.
 * `OC_RESOURCE_LABELS`: defines resource labels as a comma-seperated list of key/value pairs.
 The key MUST be seperated from the value by a single `=`. Values MAY be quoted with a single
@@ -60,9 +60,10 @@ vendors, MUST be implemented outside of the core libraries in third party or
 [census-ecosystem][census-ecosystem] repositories.
 
 ### Merging
-When new information is gained about a resource, already set labels or type fields MUST NOT
-be overwritten. Label key namespaceing SHOULD be used to prevent collisions across different
-resource detection steps.
+As different mechanisms are run to gain information about a resource, their information
+has to be merged into a single resulting resource.
+Already set labels or type fields MUST NOT be overwritten. Label key namespaceing SHOULD
+be used to prevent collisions across different resource detection steps.
 
 ### Detectors
 To make auto-detection implementations easy to use, the core resource package SHOULD define
@@ -81,6 +82,15 @@ func NewDetectorFromResource(*Resource) Detector
 // Returns a detector that runs all input detectors sequentially and merges their results.
 func ChainedDetector(...Detector) Detector
 ```
+
+### Updates
+Over the runtime of an application, resource information may change in some cases. To
+allow exporters to incorporate those changes, they SHOULD take a resource getter as a
+configuration parameter. The exporter SHOULD periodically rerun the getter and incorporate
+any changes into the data it exporters.
+If the getter returns an error, the exporter SHOULD log the error and proceed with the
+last known resource information. It SHOULD run the getter once on initialization and
+fail if it returns an error.
 
 ## Exporter translation
 A resource object MUST NOT be mutated further once it is passed to a backend-specific exporter.
@@ -101,7 +111,7 @@ For example, from a resource object
 		"cloud.google.com/zone": "eu-west2-a",
 		"cloud.google.com/project_id": "project1",
 		"cloud.google.com/gce/attributes/cluster_name": "cluster1",
-		// Populated through OpenCensus resource environment varibales.
+		// Populated through OpenCensus resource environment variables.
 		"k8s.io/namespace": "ns1",
 		"k8s.io/pod_name": "pod1",
 		"k8s.io/container_name": "container1",
@@ -124,6 +134,13 @@ resource type with well-known identifiers specific to its API:
 		"container_name": "container1",
 	},
 }
+```
+
+For another, hyopthetical, backend a simple unique identifier might be constructed instead
+by its expoerter:
+
+```
+cluster1/ns1/pod1/container1
 ```
 
 Exporter libraries MAY provide a default translation for well-known input resource types and labels.
--- a/resource/Resource.md
+++ b/resource/Resource.md
@@ -26,6 +26,8 @@ type Resource {
 }
 ```
 
+TODO(fabxc): link protobuf definition.
+
 ## Populating resources
 Resource information MAY be populated at any point between startup of the instrumented
 application and passing it to a backend-specific exporter. This explicitly includes
@@ -73,15 +75,11 @@ Example in Go:
 ```go
 type Detector func(context.Context) (*Resource, error)
 
-// Returns a detector that always returns a specific resource.
-func NewDetectorFromResource(*Resource) Detector
-
 // Returns a detector that runs all input detectors sequentially and merges their results.
 func ChainedDetector(...Detector) Detector
 ```
 
 ### Updates
-Over the runtime of an application, attributes of a resource may change in some cases.
 OpenCensus's resource representation is focused on providing static, uniquely identifying
 information and thus those mutable attributes SHOULD NOT be included in the resource
 representation.
--- a/resource/Resource.md
+++ b/resource/Resource.md
@@ -1,0 +1,133 @@
+# Resource API Overview
+The resource library primarily defines a type that captures about information about the entity
+for which stats or traces are recorded. It further provides a framework for detection of
+resource information from the environment and progressive population as signals propagate
+from the core instrumentation library to the a backend's exporter.
+
+## Resource type
+A `Resource` describes the entity for which a signal was collected through two fields:
+* `type`: a string which describes a well-known type of entity. It SHOULD be namespaced
+to avoid collisions across different environment, e.g. `kubernetes.io/container`,
+`cloud.google.com/gce/instance`. 
+* `labels`: a dictionary of labels with string keys and values that provide information
+about the entity.
+
+Type strings and label keys MAY start with a domain name and MAY be further namespaced through
+single slashes. All other characters MUST be alphanumeric. Label values MAY contain any valid
+unicode character.
+
+Implementations MAY define a `Resource` data type, constructed from the parameters above.
+Resource MAY have getters for retrieving all the information used in `Resource` definition.
+
+Example in Go:
+```go
+type Resource {
+	Type   string
+	Labels map[string]string
+}
+```
+
+## Populating resources
+Resource information MAY be populated at any point between startup of the instrumented
+application and passing it the a backend-specific exporter. This explicitly includes
+the path through future OpenCensus components such as agents or services.
+
+For example, process-identifying information may be populated through the library while
+an agent attaches further labels about the underlying VM, the cluster, or geo-location.
+
+### From environment variables
+Population of resource information from environment varibales MUST be provided by the
+core library. It provides the user with an ubiquitious way to manually provide information
+that may not be detectable automatically through available integration libraries.
+
+Two environment varibales are used:
+* `OC_RESOURCE_TYPE`: defines the resource type. Leading and trailing whitespaces are trimmed.
+* `OC_RESOURCE_LABELS`: defines resource labels as a comma-seperated list of key/value pairs.
+The key MUST be seperated from the value by a single `=`. Values MAY be quoted with a single
+leading and trailing `"`. If a values contains whitespaces, `=`, or `"` characters it MUST be
+quoted and `"` characters MUST be escaped.
+
+Population from environment variables MUST be the first applied detection process unless
+the user explicitly overwrites this behavior.
+
+### Auto-detection
+Auto-detection of resource information in specific environments, e.g. specific cloud
+vendors, MUST be implemented outside of the core libraries in third party or
+[census-ecosystem][census-ecosystem] repositories.
+
+### Merging
+When new information is gained about a resource, already set labels or type fields MUST NOT
+be overwritten. Label key namespaceing SHOULD be used to prevent collisions across different
+resource detection steps.
+
+### Detectors
+To make auto-detection implementations easy to use, the core resource package SHOULD define
+an interface to retrieve resource information. Additionally, helper functionality MAY be
+provided to effectively make use of this interface.
+The exact shape of those interfaces and helpers SHOULD be idiomatic to the respective language.
+
+Example in Go:
+
+```go
+type Detector func(context.Context) (*Resource, error)
+
+// Returns a detector that always returns a specific resource.
+func NewDetectorFromResource(*Resource) Detector
+
+// Returns a detector that runs all input detectors sequentially and merges their results.
+func ChainedDetector(...Detector) Detector
+```
+
+## Exporter translation
+A resource object MUST NOT be mutated further once it is passed to a backend-specific exporter.
+From the provided resource information, the exporter MAY transform, drop, or add information
+to build the resource identifying data type specific to its backend.
+If the passed resource does not contain sufficient information, an exporter MAY drop
+signal data entirely, if no sufficient resource information is provided to perform a correct
+write.
+
+For example, from a resource object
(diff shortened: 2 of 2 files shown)


resource/Resource.md
```diff
@@ -0,0 +1,133 @@
...
+unicode character.
+
+Implementations MAY define a `Resource` data type, constructed from the parameters above.
+Resource MAY have getters for retrieving all the information used in `Resource` definition.
+
+Example in Go:
+```go
+type Resource {
+	Type   string
+	Labels map[string]string
```
rakyll: Your implementation had Labels as Tags. Are we going with Labels?

resource/Resource.md
```diff
@@ -0,0 +1,133 @@
...
+core library. It provides the user with an ubiquitious way to manually provide information
+that may not be detectable automatically through available integration libraries.
+
+Two environment varibales are used:
+* `OC_RESOURCE_TYPE`: defines the resource type. Leading and trailing whitespaces are trimmed.
+* `OC_RESOURCE_LABELS`: defines resource labels as a comma-seperated list of key/value pairs.
+The key MUST be seperated from the value by a single `=`. Values MAY be quoted with a single
+leading and trailing `"`. If a values contains whitespaces, `=`, or `"` characters it MUST be
+quoted and `"` characters MUST be escaped.
+
```
rakyll: Can you please give examples of OC_RESOURCE_TYPE and OC_RESOURCE_LABELS to demonstrate the format?

resource/Resource.md
```diff
@@ -0,0 +1,133 @@
...
+		"location": "eu-west2-a",
+		"cluster_name": "cluster1",
+		"namespace_name": "ns1",
+		"pod_name": "pod1",
+		"container_name": "container1",
+	},
+}
+```
+
+Exporter libraries MAY provide a default translation for well-known input resource types and labels.
```
rakyll: It might be good to add a list of all known label keys here in the future.

Then, implementors can look up for the whole list and decide what they can map.

utils/MonitoredResource.md
```diff
@@ -1,84 +0,0 @@
-# Monitored Resource
```
songy23: I think this page can be left as additional information for the "Auto-detection" section.

resource/Resource.md
```diff
@@ -0,0 +1,133 @@
...
+Those would generally be based on community-supported detection integrations maintained in the
+[census-ecosystem][census-ecosystem] organisation.
+
+Additionally, exporters SHOULD provide configuration hooks for users to provide their own
+translation unless the exporter's backend does not support resources at all. For such backends,
+exporters SHOULD allow attaching converting resource labels to metric labels.
```
rghetia: not sure what you mean. what are metric labels?

nfisher:
My assumptions about Resource.md:

* this document is to outline how to associate environment related labels to spans and metrics (e.g. zone, region, pod, etc).
* The primary goal is to provide flexibility for drill-down, and fine-grained groupings in downstream tools.
* There appears to be 3 approaches being considered for associating the labels:

  1. metadata API end-points.
  2. environment variables.
  3. config file(s).

Thoughts on these approaches:

i. metadata
========

I see two mechanisms for this (although there might be others):

1. probing.
2. import at build time.

Probing doesn't feel like a great option because it makes a lot of assumptions that are liable to break (e.g. what if I'm using a custom image, what if I have a default deny policy on my machine, etc).

Importing also doesn't feel like a great option as it requires a recompilation to deploy to a new environment as I assume it would have an import that looks like this;
```golang
import _ "oc/metadata/aws"
```

I'm sure they can be made to work well but I don't think they're ideal as they feel too much like implicit magic.I can't deny their convenience in the contexts where it would just work.

ii. environment variables
==================

I think you could iterate over variables with a given prefix (e.g. `OC_LABEL_`) and use them as labels. The terms `region`, `zone`, `host`, `application`, `application_group` can map to most cloud providers:

*AWS*
```
OC_LABEL_REGION=ca-central-1
OC_LABEL_ZONE=ca-central-1a
OC_LABEL_HOST=hostname.aws
OC_LABEL_APPLICATION=my-app
OC_LABEL_APPLICATION_GROUP=my-app-asg
```
*k8s*
```
OC_LABEL_REGION=ca-central-1
OC_LABEL_ZONE=ca-central-1a
OC_LABEL_HOST=pod-instance-id1234
OC_LABEL_APPLICATION=my-app
OC_LABEL_APPLICATION_GROUP=pod
```

My assumption with the above is that OC would truncate the prefix and down-case the remaining variable.

ii. config file(s)
===========

I think simple property files could be used here with a default path relative to the binary and optionally overridden by environment variable or config option.

It's late where I am will come back to this.

resource/README.md
```diff
@@ -0,0 +1,10 @@
+# OpenCensus Library Resource Package
+This documentation serves to document the "look and feel" of the open source resource package.
+It describes they key types and overall behavior.
+
+The primary purpose of resources as a first-class concept in the core library is decoupling
```
yurishkuro: It would be useful to start with a definition of what a resource is.

resource/Resource.md
fabxc: Yes, we changed it from labels to tags. @ramonza asked elsewhere yesterday whether "labels" or "properties" would make more sense.  
Personally I don't feel strongly – there are arguments for either. Happy to change it to whatever you decide on.

fabxc:
> I saw that def but it is not precise. Specifically, is the use of term "entity" intended to generalize to more than just an application process / workload? I.e. what else an entity could be, an endpoint?

I agree that "entity" is as generic as it gets and I believe that's the best we can do in the end. Signals can refer to pretty much anything (consider IoT for example – "toothbrush" could be perfectly valid) and I don't think an instrumentation library should constrain that set artificially, especially since we can't draw any benefits from it.

fabxc:
@nfisher regarding the approaches: I think the user should be enabled to use what works for their environment. Thus the goal is to not enforce a single specific solution but to provide a thin integration package for resources – with envvars being the fallthrough, that can always be relied on.

In some environments that can mean automatic detection of resource information, which should be provided through libraries. As you said, this is unideal in the sense that it adds dependencies to the code.
That may be perfectly fine for some users – after all they'd be doing the same for the exporters they use. Both could eventually be offloaded to an OC agent/proxy that includes those dependencies once, while application binaries only send their data through a generic protocol.
The alternative is that the user has to perform the same actions the library does as a deployment step. That's certainly less magical, but also more error prone in total and a worse user experience.
Client libraries for cloud vendors also use what seems like magic to determine credentials so the user doesn't have to care about that. The auto-detection of resource information hooks into the exact same mechanisms.
In general, dumping the gathered resource information should be relatively straightforward way to debug what information was gathered or not.

For environments where auto-detection is not possible, runtime configuration is ultimately required. Envvars seemed like an easy way to integrate. 
Using prefix-based variables is indeed an option. However, they character set becomes automatically limited and namespaced label keys such as `k8s.io/pod_name` aren't possible. This is rather important as the goal is to detach gathering of resource information from interpreting it, which happens in the exporters. That ensures that exporters are truly swappable.

Config files are also a way to integrate of course. My main concern is that they make deployment more tedious (generate file, get file into the right place, permissions, ...) whereas envvars are easy to inject in most systems I'm aware of.
I think both could exist, but maybe it's worth waiting until a need for files arises?

Does that provide some background on motivation? Anything you'd like to see in the spec to clarify things?

utils/MonitoredResource.md
fabxc: My thinking was that it's more scalable to keep these docs close to the code they belong to. OTOH, with spread across languages, this repo may indeed still be the better option – maybe with separate file per detection mechanism then as those will probably grow.

resource/Resource.md
fabxc: That should be tags probably.

resource/Resource.md
fabxc: Done.

nfisher:
@fabxc agree on flexibility especially if controlled by the user as to when and how it happens. Will place a review for any further comments.

To clarify for my own understanding where would you see these labels being attached during the initialisation process?

```golang
func initTracer() {
    sdexporter, err := stackdriver.NewExporter(stackdriver.Options{ProjectID: "testing"})
    if err != nil {
        return err
    }
    trace.RegisterExporter(sdexporter)
    trace.ApplyConfig(trace.Config{DefaultSampler: trace.ProbabilitySampler(0.01)})
    http.ListenAndServe(":50030", &ochttp.Handler{})
}
```

resource/Resource.md
```diff
@@ -0,0 +1,137 @@
+# Resource API Overview
+The resource library primarily defines a type that captures information about the entity
+for which stats or traces are recorded. It further provides a framework for detection of
+resource information from the environment and progressive population as signals propagate
+from the core instrumentation library to a backend's exporter.
```
nfisher: Is there a difference between `core library` and `core instrumentation library`? Core library seems to be most consistently used throughout.

Is it worth giving an explicit definition of what the `core library` is? I assume it is any libraries provided under the Github `census-instrumentation` org which provide a common abstraction to backend exporters?

resource/Resource.md
```diff
@@ -0,0 +1,137 @@
...
+## Populating resources
+Resource information MAY be populated at any point between startup of the instrumented
+application and passing it to a backend-specific exporter. This explicitly includes
+the path through future OpenCensus components such as agents or services.
+
+For example, process-identifying information may be populated through the library while
+an agent attaches further labels about the underlying VM, the cluster, or geo-location.
+
+### From environment variables
+Population of resource information from environment varibales MUST be provided by the
```
nfisher: s/varibales/variables

resource/Resource.md
```diff
@@ -0,0 +1,137 @@
...
+```
+
+## Populating resources
+Resource information MAY be populated at any point between startup of the instrumented
+application and passing it to a backend-specific exporter. This explicitly includes
+the path through future OpenCensus components such as agents or services.
+
+For example, process-identifying information may be populated through the library while
+an agent attaches further labels about the underlying VM, the cluster, or geo-location.
+
```
nfisher: Is it worth adding a `From OC data structure` section?

resource/Resource.md
```diff
@@ -0,0 +1,137 @@
...
+// Returns a detector that runs all input detectors sequentially and merges their results.
+func ChainedDetector(...Detector) Detector
+```
+
+## Exporter translation
+A resource object MUST NOT be mutated further once it is passed to a backend-specific exporter.
+From the provided resource information, the exporter MAY transform, drop, or add information
+to build the resource identifying data type specific to its backend.
+If the passed resource does not contain sufficient information, an exporter MAY drop
```
nfisher: What is the feedback loop on dropped/ignored data? (e.g. is it a silent error, printout to stderr, returned error, etc)

My preference would be towards failing fast with a validation error at the point of library re/initialisation. Alternatively but less preferred would be a returned error at the point of capture.

resource/Resource.md
fabxc: I'd agree with the definition you gave, especially as exporters are moved to census-ecosystem.
What's the official take @bogdandrutu?

resource/Resource.md
fabxc: Mh, what do you mean specifically? Like when a resource is just passed around via internal OC APIs?

resource/Resource.md
```diff
@@ -0,0 +1,137 @@
...
+Population from environment variables MUST be the first applied detection process unless
+the user explicitly overwrites this behavior.
+
+### Auto-detection
+Auto-detection of resource information in specific environments, e.g. specific cloud
+vendors, MUST be implemented outside of the core libraries in third party or
+[census-ecosystem][census-ecosystem] repositories.
+
+### Merging
+When new information is gained about a resource, already set labels or type fields MUST NOT
```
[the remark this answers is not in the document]
fabxc: Interesting, I think the use case deviates from what I had in mind in this section.
I was mostly thinking of what happens as different auto-detection and configuration mechanisms provide more information about a resource. This would be roughly instantaneous (i.e. at program startup) though and wasn't considering the truth of a resource over time.

Sounds like this are needs clarification and we need a specific section on updating. Simplest solution would probably be to have exporters periodically re-run initialization of the monitored resource and replace the old one they are using.
Would that sound sufficient for a start?

resource/Resource.md
fabxc: Yes, as this would happen during initialization, fail-fast would be best. Will add that.
For the update case described above, emitting an error on update failures and continuing best effort with the old resource? The risk is data being a bit skewed. That seems better though than dropping all data. Especially since only small pieces of information are likely to change, which may not even be used by the exporter.

resource/Resource.md
nfisher: Labels defined in files. Main reason I arrived here to burden you with my nonsense was a tweet by @rakyll about difficulty of label discovery in k8s.

The downward API that you referenced in the deleted document seems the most reasonable/consistent approach within k8s to get those details but I'm not sure how well its constraints map to the previously outlined mechanisms (e.g. `OC_RESOURCE_LABELS` env variable and auto-detection).

fabxc: The issue with the downward API is that it can be used to set arbitrary environment variables. Currently the auto-detection OC's Stackdriver exporter does relies on a documentation example but it's not standardized. 
We could define a standard of course, which users have to adhere to. But in practice no one wants to configure every single deployment/pod with the downward API or something else. Using Kubernetes' admission control to inject this information automatically is much more feasible and powerful. And when using that, one can simply resort to the generic OC envvars proposed here without adding yet another one-off standard.

I have a working implementation for this and can hopefully release a design soon.

That said, I'd like to understand a use case where a file with information is easier than an environment variable. Even with admission control in Kubernetes as described above, implementation is a lot simpler with environment variables and a lot less likely to collide with other admission restrictions, e.g. "pods in this namespace must not mount volumes".

fabxc:
@nfisher thanks for all your comments. This is really helpful and I added some clarifications.

resource/Resource.md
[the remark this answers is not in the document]
nfisher: Given that context sounds reasonable to me.

I would be ok blocking the "ready state" of my services until all of the labels are retrieved (e.g. WaitGroup/synchronous calls) with a timeout (e.g. 500ms-1s). For me it is a preference towards consistency of labels for the lifetime of the service (where it makes sense) vs services being able to serve as quickly as possible.

fabxc:
Added back `utils/MonitoredResource.md` since this will only correspond to the independent new resource package. It can be replaced with new docs in parallel when the corresponding resource auto-detection mechanisms are added.

@nfisher I take we largely agree modulo potentially adding a file-based generic path? That could be added without conflicts at a later point.

rakyll:
I have no strong opinions on tags vs labels. But you should consider that there is a cost of introducing new terminology. We  are working on this field full time and can automatically can associate them. It is not true for an average user. @bogdandrutu, any more to add?

fabxc:
@nfisher Good point on providing the labels as debug output to a zpage. That should definitely be possible with the proposed APIs, i.e. the recommended detector functionality, will provide a function as you suggested.

resource/Resource.md
```diff
@@ -0,0 +1,154 @@
...
+for which stats or traces are recorded. It further provides a framework for detection of
+resource information from the environment and progressive population as signals propagate
+from the core instrumentation library to a backend's exporter.
+
+## Resource type
+A `Resource` describes the entity for which a signal was collected through two fields:
+* `type`: a string which describes a well-known type of entity. It SHOULD be namespaced
+to avoid collisions across different environment, e.g. `k8s.io/container`,
+`cloud.google.com/gce/instance`. 
+* `labels`: a dictionary of labels with string keys and values that provide information
```
bogdandrutu: Same comment on the strings here. Strings in go for example are encoded utf8 but in java default is utf16 what do we do here? I think it is always better to start with printable characters or a set smaller than "everything". Also if we support any character we need to do sanitization for almost all the backends, and the biggest problem is that after sanitization the keys may no longer be unique for example hence you break the protocol.

resource/Resource.md
```diff
@@ -0,0 +1,154 @@
...
+about the entity.
+
+Type strings and label keys MAY start with a domain name and MAY be further namespaced through
+single slashes. All other characters MUST be alphanumeric. Label values MAY contain any valid
+unicode character.
```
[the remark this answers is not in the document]
fabxc: Right, that should've just said "utf8" probably. But I'm fine with a subset, which it will be anyway in practice. 
Only thing I'd be worried about is, that that might not add up with backends and thus make OC restrictive without much gain.

Prometheus supports utf8 strings for label values and it hasn't been an issue yet.

resource/Resource.md
```diff
@@ -0,0 +1,154 @@
...
+	Labels map[string]string
+}
+```
+
+## Populating resources
+Resource information MAY be populated at any point between startup of the instrumented
+application and passing it to a backend-specific exporter. This explicitly includes
+the path through future OpenCensus components such as agents or services.
+
+For example, process-identifying information may be populated through the library while
```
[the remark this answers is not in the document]
fabxc: > Seems that resource can identify a VM instances or containers which means the process identity is lacking.

The resource can be as granular as the OC user likes. I don't believe this is something we should or even can solve generically.
If the user wants to export with resource labels granular enough down to a process, they can. But this will not scale for many backends. Imagine something spawns a new process very often but never runs it concurrently. If each spawning of the process created a new set of series, virtually all backends would break under the load.
I believe at some point the user has to check at which level they can ensure no collisions take place and go with that – e.g. if they know their containers only run a single process exposing signals.

Note that this proposal does not amend anything about the `<language>-<pid>@<hostname>` label attached to all metrics right now, which theoretically guarantees what you are concerned about.
Though I think it raises the problems mentioned above and that should be revisited eventually.

>  What if in a GCE VM we have multiple linux tasks how does the gce_instance resource uniquely identifies a timeseries?

Then a pure VM-based resource is the wrong choice (it almost always is). More granular labels must be provided.

> Not sure that the current proposal can completely replace that unless we try to add some mandatory fields that can get us some guarantees about uniqueness of generated timeseries.

Sounds like valuable information that exporters can make use of additional to the resource info in building the final samples they send.
Is this intended to replace the fixed `<language>-<pid>@<hostname>`  metric label?

resource/Resource.md
fabxc: Mh, does the in-memory encoding really matter? Sounds like this is up to the exporter to convert/normalize. Even if we restricted to alphanum, dashes and underscores, some backend may not like dashes and the exporter has to mutate before sending.

resource/Resource.md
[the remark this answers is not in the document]
nfisher: We (ab)use prometheus at work and have much wider range in use than ASCII. 👍 to UTF-8.

resource/Resource.md
```diff
@@ -0,0 +1,154 @@
+# Resource API Overview
```
SergeyKanzhelev: When reading "resource" there is also a scenario when you want to record operation on certain resource. Like query all traces that created, updated, read properties of, etc of a certain resource. Where resource may be some business entity. Is this an intendent use of an API?

resource/Resource.md
```diff
@@ -0,0 +1,154 @@
...
+type Resource {
+	Type   string
+	Labels map[string]string
+}
+```
+
+## Populating resources
+Resource information MAY be populated at any point between startup of the instrumented
+application and passing it to a backend-specific exporter. This explicitly includes
+the path through future OpenCensus components such as agents or services.
```
SergeyKanzhelev: Does the property like "health" be a valid field for it? If one want to mark the node as unhealthy and then healthy again? Or it's set once, change never kind of properties?

resource/Resource.md
```diff
@@ -0,0 +1,154 @@
...
+
+For example, process-identifying information may be populated through the library while
+an agent attaches further labels about the underlying VM, the cluster, or geo-location.
+
+### From environment variables
+Population of resource information from environment variables MUST be provided by the
+core library. It provides the user with an ubiquitious way to manually provide information
+that may not be detectable automatically through available integration libraries.
+
+Two environment variables are used:
```
SergeyKanzhelev: so the assumption is that there is a single resource type per process? I'm still trying to wrap my head around what is resource thus the question. Will there be option to have `admin` part of site reporting resource separately from `user` part?

resource/Resource.md
```diff
@@ -0,0 +1,154 @@
...
+* `OC_RESOURCE_TYPE`: defines the resource type. Leading and trailing whitespaces are trimmed.
+* `OC_RESOURCE_LABELS`: defines resource labels as a comma-seperated list of key/value pairs.
```
SergeyKanzhelev: BNF definition would be useful here

resource/Resource.md
```diff
@@ -0,0 +1,154 @@
...
+the user explicitly overwrites this behavior.
+
+### Auto-detection
+Auto-detection of resource information in specific environments, e.g. specific cloud
+vendors, MUST be implemented outside of the core libraries in third party or
+[census-ecosystem][census-ecosystem] repositories.
+
+### Merging
+As different mechanisms are run to gain information about a resource, their information
+has to be merged into a single resulting resource.
```
SergeyKanzhelev: automatically or on demand? It feels that there is a singleton for a resource (to my previous question). Would this singleton be per tracer? Or it should be global?

resource/Resource.md
```diff
@@ -0,0 +1,154 @@
...
+func NewDetectorFromResource(*Resource) Detector
+
+// Returns a detector that runs all input detectors sequentially and merges their results.
+func ChainedDetector(...Detector) Detector
+```
+
+### Updates
+Over the runtime of an application, resource information may change in some cases. To
+allow exporters to incorporate those changes, they SHOULD take a resource getter as a
+configuration parameter. The exporter SHOULD periodically rerun the getter and incorporate
```
SergeyKanzhelev: should there be notification mechanism for subscribers on updates? It will be useful to have if you have an exporter that needs to react on these changes. Otherwise there will be a mess of threads checking for updates from each other

resource/Resource.md
```diff
@@ -0,0 +1,154 @@
...
+
+// Returns a detector that always returns a specific resource.
+func NewDetectorFromResource(*Resource) Detector
+
+// Returns a detector that runs all input detectors sequentially and merges their results.
+func ChainedDetector(...Detector) Detector
+```
+
+### Updates
+Over the runtime of an application, resource information may change in some cases. To
```
SergeyKanzhelev: Some words on how often are required. Is health change or config update notification should be using this API?

resource/Resource.md
```diff
@@ -0,0 +1,154 @@
...
+
+Population from environment variables MUST be the first applied detection process unless
+the user explicitly overwrites this behavior.
+
+### Auto-detection
+Auto-detection of resource information in specific environments, e.g. specific cloud
+vendors, MUST be implemented outside of the core libraries in third party or
+[census-ecosystem][census-ecosystem] repositories.
+
+### Merging
```
SergeyKanzhelev: how the `type` property will be merged?

resource/Resource.md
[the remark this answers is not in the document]
fabxc: The resource does not aim describe individual service endpoints. In particular for remote endpoints, that's straight out impossible to do in the same way I'd say.

resource/Resource.md
```diff
@@ -0,0 +1,154 @@
+# Resource API Overview
+The resource library primarily defines a type that captures information about the entity
+for which stats or traces are recorded. It further provides a framework for detection of
+resource information from the environment and progressive population as signals propagate
+from the core instrumentation library to a backend's exporter.
+
+## Resource type
+A `Resource` describes the entity for which a signal was collected through two fields:
+* `type`: a string which describes a well-known type of entity. It SHOULD be namespaced
```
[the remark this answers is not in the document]
fabxc: I'd only consider doing this in the form of standardized tag keys potentially – which can be done later on when we've a better grasp of how people are using resources in different environments.
Dedicated fields will inevitably make user-side configuration more complex and the overall system harder to understand.

resource/Resource.md
fabxc: Resource is really about identifying the source of signal data and with that would be generally static. I didn't really see individual signals mutating information about their own source. 
Could you give an example?

resource/Resource.md
fabxc: In general, volatile metadata isn't helpful in identifying a resource – so I'd say no. At the same time, such data could be useful for filtering noisy/spammy signals before ingesting them.

So I think it should be allowed but exporters wouldn't actually make this data part of signal data itself.
Basically, for the near future, only static properties are of relevance IMO.

resource/Resource.md
fabxc: No, that information should generally remain part of the signal tags/labels. The resource is about identifying the signal source itself. In 90% of cases, that just means being able to identify the process through a set of tags that are meaningful within the context of the environment later, e.g. the [k8s_container](https://cloud.google.com/monitoring/api/resources#tag_k8s_container) resource in Stackdriver.
Currently, this is only possible through the `<language>-<pid>@<hostname>` label on the signal itself, which rarely makes sense to an end user later.

I get that the mental line can get kinda of blurry though.
For example, if the actual source of signals isn't able to export data itself (e.g. some embedded device) but instead a third process exports signals about N sources by proxy. Then being able to set different resources explicitly is valuable.
That would generally be an exception case that can be added by further API extensions. I'd defer that though so the design of such an extension can be guided by real-world use cases.

resource/Resource.md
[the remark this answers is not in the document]
SergeyKanzhelev: But it does try to describe a local endpoint, right? Should it be called local endpoint then?

resource/Resource.md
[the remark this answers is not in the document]
SergeyKanzhelev: I'd imagine system will want to draw applicaton map with the names taken from resource identity. In this case having `name` or `role` will be helpful as UI shouldn't worry about knowing all the different "providers" of such information.

Alernative may be to define a well-known postfix for those properties so exporter can do generic translation of labels to strongly-defined fields .

resource/Resource.md
SergeyKanzhelev: When Azure creates, modifies, etc. VMs it marks all traces performing operations on them with the VM ID. We call it resource ID =)

I can imagine the same thing applicable everywhere where business defines some resources that has a lifetime.

resource/Resource.md
SergeyKanzhelev: so resource may have property whether it's "master" or configuration "version" that is currently being used. 

Since this PR defines resources - I'd suggest to define resources lifetime and health as well.

Mostly my comment comes from the term "resource". It is so wide and general that I natirally see the desire to use it to design health and lifetime management of that resource.

resource/Resource.md
SergeyKanzhelev: So does process defines the boundary or DI container? It needs to be clearly outlined in spec.

resource/Resource.md
fabxc: That would happen explicitly during initialization/setup.

See [this PR](https://github.com/census-instrumentation/opencensus-go/pull/878) for the core library extension I proposed (Go).

With that, the only change in general would be an option when setting up the exporter, e.g.:

```go
exporter, err := stackdriver.NewExporter(stackdriver.Options{
    ResourceDetector: resource.ChainedDetector(
        resource.FromEnv,
        aws.Detect,
        azure.Detect,
        gcp.Detect,
    ),
})
```

Or, with a helper package that chains the common auto-detectors together:

```go
exporter, err := stackdriver.NewExporter(stackdriver.Options{
    ResourceDetector: auto.Detect,
})
```

resource/Resource.md
fabxc: Right now I think a periodic rerun of the detector and comparing with the previous result to detect changes will be sufficient. But there may very well be other use cases I didn't think of.

Note that the exporter-specific part of this spec is largely _SHOULD_, so there's enough flexibility to deviate/extend where necessary.

resource/Resource.md
fabxc: Per above, I'd see whether users report use cases.

Prometheus has a similar resource info model (called target info there) – different integrations update via watches/notification or just via periodic refreshes. I cannot recall a single use cases where users complained about periodic refreshes being too laggy.
It may also be interesting to see what (meta)data we added to [different integrations](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#marathon_sd_config) over time based on user demand.  

Most non-identifying information was added a) for filtering or b) because people wanted to use attach it to the signal directly.
While b) should be discourage, I think OpenCensus must allow for odd one-off use cases since it becomes most valuable, when users can use it for everything.

resource/Resource.md
fabxc: `Already set labels or type fields MUST NOT be overwritten.`

fabxc:
Thanks for the detailed feedback @SergeyKanzhelev. I tried to answer all your questions and concerns with some context.

resource/Resource.md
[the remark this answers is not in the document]
fabxc: You mean renaming `resource` to `local endpoint`? That possibly makes sense from a tracing perspective but not so much for other signals I believe.

fabxc: So the traces get tagged with the VM they are coming from?
That would essentially be exactly what this proposal aims at – just that machine identifiers are often to coarse and in the case of metrics, basically always.

resource/Resource.md
fabxc: I think whether a resource has health or lifetime info to begin with really depends on the resource. If it has this information, a library implementing a resource detector can easily provide this info through dedicated labels.

Though, as hinted at in other comments, I think things should mostly be constrained to static resource information. Backends can pull into volatile metadata at query time from other sources.

resource/Resource.md
fabxc: My opinion differs on this a bit. I think OC should provide clear docs around best practices – we can probably also some SHOULD/SHOULD NOT sentences to the spec.

But a clear line is not possible with this in my experience. There'll always be things we'd clearly see as "MUST NOT do this" that people might have a valid one-off use case for. If we don't enable those, even if they don't yield ideal results, OC becomes significantly less valuable for those users.
If 1 out of 20 teams in an org cannot use the instrumentation system of choice, the negative impact is far greater than 5%.

resource/Resource.md
SergeyKanzhelev: no, with the VMs they operate on. VM here is a customer VM created via call to Azure. Sorry for confusion

fabxc: Oh I see, the Azure API itself here being what's instrumented.

I think in that case the VM ID would just be part of the signal tag itself, i.e. a tag of trace rather than a tag of the resource of the trace.

resource/Resource.md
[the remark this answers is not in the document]
fabxc: I see what you are getting at. I agree that in the system where the data ends up, one wants to have more structure to enable certain workflows.

This proposal does not extend as far as telling backend system to have a resource presentation in form of a type string and tag map. Instead, the specific exporters should convert the rather structure-less resource notion proposed here into something more structured.
For Stackdriver for example, the exporter is configured with a mapping that converts known resource type and tag keys to Stackdriver Monitored Resources (e.g. [k8s_container](https://cloud.google.com/monitoring/api/resources#tag_k8s_container)) which have fixed fields.
Exporters will by default do a mapping of tags set by well-known resource detectors.

resource/Resource.md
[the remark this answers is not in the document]
SergeyKanzhelev: what other signals it wouldn't make sense for?

resource/Resource.md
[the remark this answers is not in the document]
SergeyKanzhelev: Makes sense. What I'm trying to accomplish is some hint from resource-auto-discovery logic to exporter on what is what. So perhaps some postfix in names that help determine whether this tag is about unique ID of instance or logical group of instances (role). Or have those two on the level with `type`. Otherwise every exporter will need to know about every resource auto-detector.

resource/Resource.md
SergeyKanzhelev: this comment comes from the very short intro of this document. Maybe make it more explicit what is resource.

resource/Resource.md
SergeyKanzhelev: health and lifetime tracking requires more responsive mechanism to detect changes that pulling. If this is a scenario - it should be taken into consideration

resource/Resource.md
SergeyKanzhelev: heatlth and lifetime management are scenarios that might require more immidiate update of properties.

SergeyKanzhelev:
@fabxc thank you! I understand it much better now. Since it's all in-proc I believe more responsive update notification mechanism may be beneficial for health and lifetime tracking of the resource. Also some strongly typed fields might be helpful to simplify writing exporters. Do you think it's something this spec can accomodate?

resource/Resource.md
bogdandrutu: I think core library is better.

resource/Resource.md
bogdandrutu: I think the in-memory representation matters for one reason: we are giving people an option to set these strings via environment variables so we have to define what a valid input it is for the key/value strings.

SergeyKanzhelev:
@bogdandrutu it makes sense. If you think of resources as a way to design this type of hierarchy - there are also layers like role (set of endpoints - like "health", "admin") and endpoint itself.

It also calls to association of an individual span with the resource. Not for exporter to pull data from resources. Is it what you are thinking about?

resource/Resource.md
bogdandrutu: @SergeyKanzhelev I would like to understand the "heatlth and lifetime management are scenarios". Can you give me more details about what you have in mind.

My personal opinion is to keep simplicity and actually not support runtime config for these in the library at all, unless we have a clear use-case which I miss here, but I can be convinced if you have one in mind.

resource/Resource.md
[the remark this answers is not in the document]
fabxc: A process doing file process can still expose metrics about its internal state without doing any request/response flows at all. Same for logs/events should they become part of OC at some point.

resource/Resource.md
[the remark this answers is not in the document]
fabxc: I'd expect the number of active detectors in a given environment to be very low so that configuration isn't a particularly big problem. Plus, if most common environments are supported by default, only a small fraction of users has to do configuration at all.

Coming up with a canonical set of schemas (in the fashion of [schema.org](https://schema.org)) will eat lots of time and is hard to get right. If it only relieves a few lines if configuration for a minority of users, I'm not sure that's worth it.

Again Prometheus as a data point – the provided data has even weaker structure and no default behavior at all, i.e. all users have to write rather awkward configuration. Regardless, they largely consider this to be a feature.

resource/Resource.md
fabxc: 👍

resource/Resource.md
SergeyKanzhelev: The key of heatlth and lifetime management is that different modules may contribute into the common state and make decisions based on it. Like if CPU is too high - make sampling more aggressive and report yellow state of the app on this node. Or if exporter failed to upload telemetry with many retries - switch an app to red state and potentially node out of rotation. Or when app was swapped into production from staging - start collecting more metrics. 

I agree with you that resource concept like proposed can be implemented if there is a way to set global attributes like proposed in https://github.com/census-instrumentation/opencensus-specs/pull/167. Maybe make it more elaborate and implement it via callbacks to resource auto-discovery modules (without calling them that).

So static and almost static informaiton is not that hard to implement without introducing new concepts (except global attributes or attributes callbacks).

fabxc:
Added some clarifications around the allowed character set (it's more restrictive now and matches tags).

@SergeyKanzhelev I changed the section on updates following a discussion with @mayurkale22 and @bogdandrutu to discourage usage of any mutable attributes in resource labels. On the one hand this is definitely the safe route and we can always lift restrictions based on the future feedback. On the other hand, it seems like mutable runtime information like health serves a quite different purpose from the resource feature discussed here.  
Thus it's probably a good idea to think about it as an orthogonal feature in OC as well. Would you like to create a separate feature request issue so we can discuss use cases and potential implementation there?

resource/README.md
bogdandrutu: Please resolve this. I think it is important to specify what a "resource" is, maybe include examples.

resource/Resource.md
```diff
@@ -0,0 +1,150 @@
...
+Type, label keys, and label values MUST contain only printable ASCII (codes between 32
+and 126, inclusive) and not exceed 256 characters.
+Type and label keys MUST have a length greater than zero. They SHOULD start with a domain
+name and separate hierarchies with `/` characters, e.g. `k8s.io/namespace/name`.
+
+Implementations MAY define a `Resource` data type, constructed from the parameters above.
+`Resource` MUST have getters for retrieving all the information used in `Resource` definition.
+
+Example in Go:
+```go
```
bogdandrutu: Maybe add a TODO to add a link to the proto definition.

resource/Resource.md
```diff
@@ -0,0 +1,150 @@
...
+provided to effectively make use of this interface.
+The exact shape of those interfaces and helpers SHOULD be idiomatic to the respective language.
+
+Example in Go:
+
+```go
+type Detector func(context.Context) (*Resource, error)
+
+// Returns a detector that always returns a specific resource.
+func NewDetectorFromResource(*Resource) Detector
```
bogdandrutu: I think this is more like a helper method. Not necessary to have it in the specs.

resource/Resource.md
```diff
@@ -0,0 +1,150 @@
...
+
+// Returns a detector that runs all input detectors sequentially and merges their results.
+func ChainedDetector(...Detector) Detector
```
bogdandrutu: Chained does not sound very good to me. Maybe ResolveResourceFromDetector or maybe you have better suggestions.

resource/Resource.md
```diff
@@ -0,0 +1,150 @@
...
+```
+
+### Updates
+Over the runtime of an application, attributes of a resource may change in some cases.
```
bogdandrutu: Here this sentence implies we support dynamic changes of the resource fields. Or maybe I am just wrong in understanding this.

resource/Resource.md
fabxc: Mh, it doesn't resolve a resource though but rather just turns a list of Detectors into a single Detector again that has to be called first.

Best alternative I can think of is `CombinedDetector`.

fabxc:
@SergeyKanzhelev did you have a chance to take a look at this?

SergeyKanzhelev:
@fabxc thank you for addressing feedback. Looks good to me. Sorry for delay

PR #167, opened 2018-08-28
set top level attribute(s) for the entire trace

mayurkale22:


basvanbeek:
I agree with @bogdandrutu that can better live in the exporter properties.

Where I expect default attributes to be of use is in the scope of specific instrumentation (ex. http transport middleware, sql middleware, etc.) where the attributes would describe things like http client version, sql driver, etc.)

mayurkale22:
@bogdandrutu 
Implementation should provide a way to set global attributes during any exporter register process and provide a way to override/skip default attributes per-span (something like skipGlobalAttributes(true):java)

kazegusuri:
+1 for this feature
I want to populate system/runtime information into Span. I think it's not exporter specific attributes.

PR #175, opened 2018-09-12
Fix broken link

mayurkale22:
The property name change is as per, https://cloud.google.com/monitoring/api/resources#tag_gce_instance

PR #178, opened 2018-09-17
utils/MonitoredResource: fix broken GCPMetadataJava with permalink

odeke-em:
Use a permanent-link "permalink" to reference GCPMetadataJava
since the previous one pointed to a resource that was at
the mercy of the maintainers, and in deed they moved it to
`*/google-cloud-clients/google-cloud-core/*` and that
broke this reference.


mayurkale22:
@odeke-em https://github.com/census-instrumentation/opencensus-specs/pull/175 same PR.

odeke-em:
@mayurkale22 not really though, that PR uses a non-permalink this one uses a permalink as the title says :)

PR #179, opened 2018-09-19
Add a document describing exemplars

ramonza:


ramonza:
This has been partially implemented already but I would like to describe exemplars officially here.

stats/Exemplars.md
```diff
@@ -0,0 +1,46 @@
...
+describe the context in which the exemplar was recored.
+
+Well-known attachment keys
+---
+
+We define the following types of attachments:
+
+1. `TraceID`: hex encoded trace ID
+2. `SpanID`: hex encoded span ID
+3. `TraceOptions`: decimal trace options value
```
bogdandrutu: Can we try to be consistent and use hex here as well?

stats/Exemplars.md
```diff
@@ -0,0 +1,46 @@
...
+
+Exemplars my have zero or more attachments. Attachments are key-value pairs that
+describe the context in which the exemplar was recored.
+
+Well-known attachment keys
+---
+
+We define the following types of attachments:
+
+1. `TraceID`: hex encoded trace ID
```
bogdandrutu: I would say lower base16 because we say that the trace ID is byte[16] not a number. Also that ensures this has 32 hex characters all the time.

stats/Exemplars.md
```diff
@@ -0,0 +1,46 @@
...
+If `TraceID` and `SpanID` are provided, the associated trace should be sampled.
+
+Retention in views
+---
+
+Libraries should attempt to retain and export at least once exemplar per bucket per reporting
+period for distribution-type views. Libraries should not retain an unbounded number
+of exemplars.
+
+Libraries should prefer exemplars with sampled traces associated with them. After considering
```
bogdandrutu: Maybe refer to what sampled means.

ramonza:
@bogdandrutu addressed comments

ramonza:
cc @songy23 @dinooliva

stats/Exemplars.md
```diff
@@ -0,0 +1,48 @@
+Exemplars
+===
```
SergeyKanzhelev: most of documents are using `#` for headers. Can you please use the same?

stats/Exemplars.md
```diff
@@ -0,0 +1,48 @@
...
+
+Exemplars are example data points for aggregated data. In particular, for distribution-type
```
SergeyKanzhelev: Is this document proposes new API or event schema? Or it is simply a terminology description?

stats/Exemplars.md
```diff
@@ -0,0 +1,48 @@
...
+describe the context in which the exemplar was recored.
+
+Well-known attachment keys
+---
+
+We define the following types of attachments:
+
+1. `TraceID`: lower-case base16 encoded trace ID
+2. `SpanID`:  lower-case base16 encoded span ID
+3. `TraceOptions`: lower-case base16 encoded trace options
```
SergeyKanzhelev: this is kind of irrelevant as you asking to sample it anyway. Or this document tells to override the sampling flag for downstream services as well?

stats/Exemplars.md
```diff
@@ -0,0 +1,48 @@
...
+4. `Tag:KEY`: tags that were active at the time of recording, regardless of whether
```
SergeyKanzhelev: what about `tracestate`?

stats/Exemplars.md
```diff
@@ -0,0 +1,48 @@
...
+Libraries should prefer retaining exemplars with
+[sampled traces](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/Sampling.md)
+associated with them.
+After considering associated trace, libraries should prefer exemplars with more attachments
+over exemplars with fewer or no attachments.
+
+Exporting
+---
+
+The interpretation of attachments is up to the exporter. Currently, only Stackdriver is known
```
SergeyKanzhelev: Should it be other way around - new properties for a span that indicate which metric this span is example of? Or should it be both?

stats/Exemplars.md
ramonza: There's already protobuf for this and an implementation in Java and I'm working on a Go implementation.

stats/Exemplars.md
ramonza: Yeah agreed, I will remove it for now.

stats/Exemplars.md
ramonza: Not sure how to handle it. For now I recommend we leave it out.

ramonza: To clarify, I don't really have good knowledge on what tracestate is - open to suggestions on alternative ways to handle it here.

stats/Exemplars.md
ramonza: Yeah I really like that approach since it works with existing trace backends. It has the same goals as trace exemplars but it's a different way to achieve those goals so I'm not sure it belongs in this doc. Maybe you should add something to one of the tracing docs to recommend that we add properties/annotations to spans like this?

stats/Exemplars.md
SergeyKanzhelev: and also make a note whether this document asks for sampling individual span in or everything downstream as well

stats/Exemplars.md
SergeyKanzhelev: do you know if this is part of OpenMetrics charter as well? If so - perhaps it should be less about traces and more about metrics

SergeyKanzhelev: + @mtwo

stats/Exemplars.md
SergeyKanzhelev: I'd say there should be a callback to allow people to extract more, especially from `tracestate`

stats/Exemplars.md
bogdandrutu: Yes it is part of OpenMetrics, they have support for exemplars in their definition.

ramonza: It's not only tracing-specific: this doc also proposes we add the tags as attachments.

stats/Exemplars.md
ramonza: I will add: 
```
Implementations should provide an extensible mechanism to extract custom attachments
from the context.
```

ramonza:
@SergeyKanzhelev @bogdandrutu PTAL

SergeyKanzhelev:
I still want to see more explicit reference on Open Metrics exemplars in the document introduction. Otherwise LGTM.

ramonza:
@SergeyKanzhelev I could not find anything to link to, happy to add any link you think is relevant.

SergeyKanzhelev:
@ramonza me too =) I hopes Open Metrics moved further forward. LGTM as is than

bogdandrutu:
I have access to the current OpenMetrics draft and it has exemplars. I personally cannot make it public because the working group prefers to keep documents private for the moment.

Issue #123, opened 2018-06-15
Add support for correlating traces and logs.


sebright:
One approach is to add trace IDs, span IDs, or other fields from the current span to log entries that are created with existing logging frameworks.  I am currently experimenting with this feature in opencensus-java (https://github.com/census-instrumentation/opencensus-java/pull/1212), but I think we should design a feature that can be supported across multiple languages and include it in opencensus-specs.

sebright:
I added two experimental log correlation libraries to opencensus-java:

Log4j: https://github.com/census-instrumentation/opencensus-java/tree/master/contrib/log_correlation/log4j

Stackdriver (google-cloud-logging): https://github.com/census-instrumentation/opencensus-java/tree/master/contrib/log_correlation/stackdriver

sebright:
One issue that came up was deciding how to name the keys for the tracing data, when the logging framework supports adding arbitrary key-value pairs to log entries.  We should probably standardize the key names.  See the discussion at https://github.com/census-instrumentation/opencensus-java/pull/1371#discussion_r211440499.

The Log4j log correlation library currently uses the keys `openCensusTraceId`, `openCensusSpanId`, and `openCensusTraceSampled`.  There is also a feature request to allow customizing the key names: https://github.com/census-instrumentation/opencensus-java/issues/1389.

sebright:
Closing, since #181 added log correlation, and #195 covers key naming.

PR #181, opened 2018-09-20
Add specification of log correlation for tracing (issue #123).

sebright:
The specification only covers aspects of log correlation that are likely to be
shared by log correlation implementations for multiple languages and logging
frameworks.  It is based on the experimental log correlation libraries in
opencensus-java:

https://github.com/census-instrumentation/opencensus-java/tree/master/contrib/log_correlation/stackdriver
https://github.com/census-instrumentation/opencensus-java/tree/master/contrib/log_correlation/log4j2

/cc @ramonza @odeke-em @rghetia


trace/LogCorrelation.md
```diff
@@ -0,0 +1,63 @@
+# Log Correlation
+
+Log correlation is a feature that inserts information about the current span into log entries
+created by existing logging frameworks.  The feature can be used to add more context to log entries,
+filter log entries by trace ID, or show log entries as annotations on a trace.
```
rakyll: "log entries as annotations on a trace"

This is a possibility but do we want to encourage this usage? I think we should keep it out.

trace/LogCorrelation.md
```diff
@@ -0,0 +1,63 @@
...
+The span ID of the current span.  See [Span.md#spanid](Span.md#spanid).
+
+### Sampling Decision
+
+The sampling bit of the current span, as a boolean.  See
+[Span.md#supported-bits](Span.md#supported-bits).
+
+## String format for tracing data
+
+The logging framework may require the pieces of tracing data to be converted to strings.  In that
```
rakyll: Tracing data or identifiers?

trace/LogCorrelation.md
```diff
@@ -0,0 +1,63 @@
...
+case, the log correlation implementation should format the trace ID and span ID as lowercase base 16
+and format the sampling decision as "true" or "false".
+
+## Key names for tracing data
+
+Some logging frameworks allow the insertion of arbitrary key-value pairs into log entries.  When
+a log correlation implementation inserts tracing data by that method, the key names should be
+"opencensusTraceId", "opencensusSpanId", and "opencensusTraceSampled" by default.  The log
```
rakyll: These are very mouthful key names.

OTOH, underscore is often preferred over camel case in log labels. We might want to reconsider these keys.

sebright: > These are very mouthful key names.

There was some discussion about short vs long key names in https://github.com/census-instrumentation/opencensus-java/pull/1371#discussion_r211440499.

I thought that it would be better to use long key names by default, to avoid conflicting with other tracing libraries' key names.  Users could then override the key names if they wanted to purposefully use the same key names for, say, Brave and OpenCensus.

> OTOH, underscore is often preferred over camel case in log labels. We might want to reconsider these keys.

I'm happy to change the key names if underscores are more common.  Where have you seen underscores used?

I saw camel case logging context key names in Brave, Wingtips, and Spring Cloud Sleuth and camel case tracing-related log entry fields in Stackdriver (https://cloud.google.com/logging/docs/agent/configuration#special_fields_in_structured_payloads).  The three libraries are all Java, though, so it could easily be Java specific.

trace/LogCorrelation.md
sebright: Do you mean it should say "The logging framework may require the tracing data or identifiers to be converted to strings"?

trace/LogCorrelation.md
rakyll: It could be a language-community preference. I saw it on Stackdriver Logging, zap and logrus.

https://github.com/uber-go/zap
https://github.com/sirupsen/logrus

ramonza: In the unusual case of conflict, users can always override the defaults.

trace/LogCorrelation.md
adriancole: I think the assumption that this would be used to decorate traces might be faulty. it would be more likely to have correlated queries especially as tracing uis are not as good at filtering logs (which can be extremely volumnous) than logging uis.

adriancole: so I agree we should not anchor people with the idea that this will result in log entries as span annotations. eg leave it out or rename to language that is correlation in nature vs the word annotation which is historically the name of a part of a span

trace/LogCorrelation.md
```diff
@@ -0,0 +1,70 @@
...
+
+A log correlation implementation should look up tracing data from the span that is current at the
+point of the log statement.  See
+[Span.md#how-span-interacts-with-context](Span.md#how-span-interacts-with-context) for the
+definition of the current span.
+
+## Tracing data to include in log entries
+
+A log correlation implementation should insert the following pieces of tracing data from the current
+span context into each log entry:
```
adriancole: make available in the logging entry. eg in some systems it might not literally be a copy rather sharing in context.

trace/LogCorrelation.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Trace ID
+
+The trace ID of the current span.  See [Span.md#traceid](Span.md#traceid).
+
+### Span ID
```
adriancole: note it is common for some tooling to attempt to reverse engineer spans from log entries. when this is the case the parent id is also put in the logging context

trace/LogCorrelation.md
adriancole: I think we should defer to language defaults. it would be weird to not use lower camel in most java logging formats (though not all). so we can make examples in a case format and caveat that appropriate use will occur when applied to a logging framework

trace/LogCorrelation.md
```diff
@@ -0,0 +1,70 @@
...
+and format the sampling decision as "true" or "false".
+
+## Key names for tracing data
+
+Some logging frameworks allow the insertion of arbitrary key-value pairs into log entries.  When
+a log correlation implementation inserts tracing data by that method, the key names should be
+"opencensusTraceId", "opencensusSpanId", and "opencensusTraceSampled" by default.  The log
+correlation implementation may allow the user to override the tracing data key names.
+
+## Deciding when to add tracing data to a log entry
```
adriancole: I think this feature is a bit speculative. I would leave it out as I have only seen this practice here. usually it isnt selective.

trace/LogCorrelation.md
```diff
@@ -0,0 +1,70 @@
...
+### Span ID
+
+The span ID of the current span.  See [Span.md#spanid](Span.md#spanid).
+
+### Sampling Decision
+
+The sampling bit of the current span, as a boolean.  See
+[Span.md#supported-bits](Span.md#supported-bits).
+
+## String format for tracing data
```
SergeyKanzhelev: I suggest to add a section for "Other fields from tracestate" or something like this. So vendors-specific key/value pairs can be added to the logs.

trace/LogCorrelation.md
bogdandrutu: Maybe add a todo to consider this. Until we have a clear request it is not worth doing it.

trace/LogCorrelation.md
```diff
@@ -0,0 +1,63 @@
+# Log Correlation
+
+Log correlation is a feature that inserts information about the current span into log entries
```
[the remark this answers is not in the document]
bogdandrutu: +1 for @ramonza.

trace/LogCorrelation.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Trace ID
+
+The trace ID of the current span.  See [Span.md#traceid](Span.md#traceid).
+
+### Span ID
+
+The span ID of the current span.  See [Span.md#spanid](Span.md#spanid).
+
+### Sampling Decision
```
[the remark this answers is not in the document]
bogdandrutu: @SergeyKanzhelev do you want to write a one pager explaining the samplingScore, how can be used, why is important to be propagated, etc.

For the moment I would not mention it here until we define it. We can leave a TODO to say when decide on sampling score then consider to add it here.

trace/LogCorrelation.md
sebright: Done

trace/LogCorrelation.md
bogdandrutu: I think it worth mentioning the Tracestate here. Probably every integration should offer a callback mechanism to extract and log data from Tracestate.

sebright: How about I add a TODO and open an issue for this part?  I think it will require more discussion, and this PR already has a lot of comments.

trace/LogCorrelation.md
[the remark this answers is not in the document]
sebright: I added a TODO.

trace/LogCorrelation.md
bogdandrutu: I think supporting per language name is not the best option:
1. Usually the log backends are not per language. So the backend has to understand all of these combinations.
2. I agree that size matters here so probably shorter is better (less data written).
3. I understand that for Java a key like "oc_trace_id" is probably not a good fit. For the moment we have only Java and Go here. Can we get some input from other languages?

Even if we have multiple keys we need to document them here for each language.

trace/LogCorrelation.md
sebright: I added a TODO for this feature.

sebright:
I reverted the changes to the key names for now, and I'm planning to open an issue for deciding on the key names separately.

trace/LogCorrelation.md
sebright: I opened an issue for continuing the discussion: #195

sebright:
I labeled the specification as a draft and linked to the issue about deciding on key names.  I'll merge it now.

trace/LogCorrelation.md
sebright: I opened #196.

sebright:
I think I addressed all of the comments by updating the PR, adding TODOs, or opening issues, but please let me know if I missed anything.  Thanks for the reviews!

PR #183, opened 2018-09-23
Http out test cases

SergeyKanzhelev:
This is how http out test cases may look like. There are examples on how to use them in C#: https://github.com/census-instrumentation/opencensus-csharp/pull/43 and in Go: https://github.com/census-instrumentation/opencensus-go/pull/928

### DIFF
--- a/trace/HTTP.md
+++ b/trace/HTTP.md
@@ -103,18 +103,20 @@ Implementations SHOULD create message event when response size is determined.
 ## Attributes
 
 Implementations SHOULD set the following attributes on the client and server spans. For a server,
- request represents the incoming request. For a client, request represents the outgoing request.
+request represents the incoming request. For a client, request represents the outgoing request.
 
-All attributes are optional.
+All attributes are optional, but collector should make the best effort
+to collect those.
 
 | Attribute name            | Description                 | Example value                   |
 |---------------------------|-----------------------------|---------------------------------|
-| "http.host"               | Request URL host            | "example.com:779"               |
-| "http.method"             | Request URL method          | "GET"                           |
-| "http.path"               | Request URL path            | "/users/25f4c31d"               |
-| "http.route"              | Matched request URL route   | "/users/:userID"                |
-| "http.user_agent"         | Request user-agent. Do not inject attribute if user-agent is empty. | "HTTPClient/1.2" |
+| "http.host"               | Request URL host            | `example.com:779`               |
+| "http.method"             | Request URL method          | `GET`                           |
+| "http.path"               | Request URL path. If empty - set to `/`            | `/users/25f4c31d`  |
+| "http.route"              | Matched request URL route   | `/users/:userID`                |
+| "http.user_agent"         | Request user-agent. Do not inject attribute if user-agent is empty. | `HTTPClient/1.2` |
 | "http.status_code"        | Response status code        | 200                             |
+| "http.url"                | Response status code        | 200                             |
 
 Exporters should always export the collected attributes. Exporters should map the collected 
 attributes to backend's known attributes/labels.
@@ -130,6 +132,13 @@ known attributes/labels on supported tracing backends.
 | "http.route"              | "http.route"       | "http.route"       | "/http/route"             |
 | "http.user_agent"         | "http.user_agent"  | "http.user_agent"  | "/http/user_agent"        |
 | "http.status_code"        | "http.status_code" | "http.status_code" | "/http/status_code"       |
+| "http.url"                | "http.url"         | "http.url"         | "/http/url"               |
+
+References:
+
+- [Stackdriver Trace
+  label](https://cloud.google.com/trace/docs/reference/v1/rest/v1/projects.traces)
+- [Jaeger/Open Tracing](https://github.com/opentracing/specification/blob/master/semantic_conventions.md)
 
 ## Test Cases
 
--- a/trace/HTTP.md
+++ b/trace/HTTP.md
@@ -108,17 +108,17 @@ request represents the incoming request. For a client, request represents the ou
 All attributes are optional, but collector should make the best effort
 to collect those.
 
-| Attribute name            | Description                 | Example value                   |
-|---------------------------|-----------------------------|---------------------------------|
-| "http.host"               | Request URL host            | `example.com:779`               |
-| "http.method"             | Request URL method          | `GET`                           |
-| "http.path"               | Request URL path. If empty - set to `/`            | `/users/25f4c31d`  |
-| "http.route"              | Matched request URL route   | `/users/:userID`                |
-| "http.user_agent"         | Request user-agent. Do not inject attribute if user-agent is empty. | `HTTPClient/1.2` |
-| "http.status_code"        | Response status code        | 200                             |
-| "http.url"                | Response status code        | 200                             |
+| Attribute name            | Description                 | Type   |Example value              |
+|---------------------------|-----------------------------|--------|---------------------------|
+| "http.host"               | Request URL host            | string | `example.com:779`         |
+| "http.method"             | Request URL method          | string | `GET`                     |
+| "http.path"               | Request URL path. If empty - set to `/` | `/users/25f4c31d`  |
+| "http.route"              | Matched request URL route   | string | `/users/:userID`          |
+| "http.user_agent"         | Request user-agent. Do not inject attribute if user-agent is empty. | string | `HTTPClient/1.2` |
+| "http.status_code"        | Response status code        | int64  | `200`                     |
+| "http.url"                | Absolute request URL        | string | `https://example.com:779/path/12314/?q=ddds#123` |
 
-Exporters should always export the collected attributes. Exporters should map the collected 
+Exporters should always export the collected attributes. Exporters should map the collected
 attributes to backend's known attributes/labels.
 
 The following table summarizes how OpenCensus attributes maps to the
--- a/trace/HTTP.md
+++ b/trace/HTTP.md
@@ -139,6 +139,7 @@ References:
 - [Stackdriver Trace
   label](https://cloud.google.com/trace/docs/reference/v1/rest/v1/projects.traces)
 - [Jaeger/Open Tracing](https://github.com/opentracing/specification/blob/master/semantic_conventions.md)
+- [Zipkin](https://github.com/openzipkin/zipkin-api/blob/master/thrift/zipkinCore.thrift)
 
 ## Test Cases
 
--- a/trace/HTTP.md
+++ b/trace/HTTP.md
@@ -72,7 +72,7 @@ response status line (if available).
 ### Client errors for client HTTP calls
 
 There are a number of client errors when trying to access http endpoint. Here
-are examples of mapping those to the Open Census status codes.
+are examples of mapping those to the OpenCensus status codes.
 
 | Client error                 | Trace status code     |
 |------------------------------|-----------------------|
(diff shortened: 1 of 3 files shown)


SergeyKanzhelev:
CC: @rghetia

SergeyKanzhelev:
@rghetia @lmolkova python SDK also collects `http.url` property. Which was not in spec. But will be very valuable as now we lose complete query string.


Will it be better to add `http.url` or `http.query_string` from your opinion?

rghetia:
http.query_string will be useful.

lmolkova:
Spec mentions that `All attributes are optional.`. I think we should change this to something like: http instrumentation must make the best effort to report the following attributes (whatever we decide is needed). Otherwise it seems instrumentation can do anything.

I also wonder how far we can go modifying existing instrumentations that report wrong things. E.g. 
in python on 'httplib' level `http.url` is actually relative url and it's different on 'requests' level. Changing it may break some UX.

If we use `http.url` and make everyone use absolute url, it simplifies things a bit: 
* All properties are in the same place: e.g. schema: now you won't be able to tell if you hit https://google.com or http://google.com. Port is -1 in either case (at least in python) and knowing port is not enough. Is schema important? As a user, you may find it interesting that you service calls http endpoint instead of secure one.

* it's probably more efficient in terms of memory/bandwidth
* it's less ambiguous: what if some instrumentation sets host, but not path or query: it complicates exporter behavior and required to create some defaults. E.g. in App Insights we use full urls and it's a nightmare to construct a full url from pieces each of which is potentially missing.

So, I prefer 'http.url' and a statement that it's an absolute url.

rghetia:
is the expectation that if url is set then host and path will not be set? Otherwise it is duplication of data.

SergeyKanzhelev:
With `url` there is some duplication of data. With `query_string` we are loosing information on protocol and it's unclear whether named anchor should be included. 

I think `url` is a cleaner solution. It is also aligned with zipkin and jaeger. My understanding why duplication is needed is to allow easier aggregation by those attributes - without parsing.

My vote is also for `http.url`. Other opinions?

trace/HTTP.md
```diff
@@ -90,18 +103,20 @@ Implementations SHOULD create message event when response size is determined.
...
-| "http.path"               | Request URL path            | "/users/25f4c31d"               |
-| "http.route"              | Matched request URL route   | "/users/:userID"                |
-| "http.user_agent"         | Request user-agent          | "HTTPClient/1.2"                |
+| "http.host"               | Request URL host            | `example.com:779`               |
+| "http.method"             | Request URL method          | `GET`                           |
+| "http.path"               | Request URL path. If empty - set to `/`            | `/users/25f4c31d`  |
+| "http.route"              | Matched request URL route   | `/users/:userID`                |
+| "http.user_agent"         | Request user-agent. Do not inject attribute if user-agent is empty. | `HTTPClient/1.2` |
 | "http.status_code"        | Response status code        | 200                             |
+| "http.url"                | Response status code        | 200                             |
```
lmolkova: Please fix description and example

trace/HTTP.md
```diff
@@ -90,18 +103,20 @@ Implementations SHOULD create message event when response size is determined.
 ## Attributes
 
 Implementations SHOULD set the following attributes on the client and server spans. For a server,
- request represents the incoming request. For a client, request represents the outgoing request.
+request represents the incoming request. For a client, request represents the outgoing request.
 
-All attributes are optional.
+All attributes are optional, but collector should make the best effort
+to collect those.
```
lmolkova: can we ask to provide wither url or host and path? adding all of them seems redundant

trace/HTTP.md
SergeyKanzhelev: oh, sorry =)

trace/HTTP.md
SergeyKanzhelev: it is redundant. But it may help to set up metrics aggregation and build exporters easier. `Path` should be calculated anyway to construct `name`. `Host` is often used for aggregations and it will be extracted by exporter in most cases. So I'd suggest to keep all of them together. What is your biggest concern here - wire size or performance?

lmolkova: I believe that Url parsing is relatively cheap and we are not going to gain much by duplication and avoiding parsing.

 My biggest concern is ambiguity and inconsistency.   I think our goal is to have all instrumentations to report the same data (when it's feasible). If we provide two different ways to report the same information, there will be some inconsistency as instrumentations would use whatever is more convenient, especially when there is obvious redundancy in the data.
Cost of changing the instrumentation in the backward compatible way is really high and is not justifiable, so we'll live with slightly different instrumentations and all kinds of workarounds on backends forever.

So, I'd prefer we have just one and straightforward way to set all http related attributes, but it would be great to hear what @adriancole and Jaeger and stack driver people think about it as they seem to rely on current attribute spec.

adriancole: I think what is collected and what is exported is the key issue. for example as a programming structure it matters less as you can derive host from url etc. some libraries you cannot get url (as you dont know scheme) so it is fake which might be ok.

problem really is when we simultaneously use same thing to talk about structs and then what is exported. I would prefer a policy where something claims the tags it needs. communicating this might be tricky now that collection tier is in scope of the project. for example who owns that config? does collector push data policy to the library?

SergeyKanzhelev:
@adriancole I'm not sure we need this mapping table at all in this spec. I just included references for easier troubleshooting.

trace/HTTP.md
SergeyKanzhelev: I understand you are advocating to collect all proposed. Exporter may decide to cut some?

It is very useful to have a min bar for which properties one should expect be collected for http.

trace/HTTP.md
```diff
@@ -117,3 +132,49 @@ known attributes/labels on supported tracing backends.
...
+    "http.host": "{host}:{port}",
+    "http.status_code": "200",
+    "http.user_agent": "test-user-agent"
+}
+}
+```
+
+Where `name` is the name of the test case. Properties `method`, `url` and
+`headers` collection represents the outgoing call. The field `responseCode`
+describes the response code status code.
```
rghetia: nit:s/code status/status/

trace/http-out-test-cases.json
```diff
@@ -0,0 +1,274 @@
...
+      "spanKind": "Client",
+      "spanAttributes": {
+        "http.path": "/",
+        "http.method": "GET",
+        "http.host": "sdlfaldfjalkdfjlkajdflkajlsdjf.sdlkjafsdjfalfadslkf.com",
+        "http.url": "https://sdlfaldfjalkdfjlkajdflkajlsdjf.sdlkjafsdjfalfadslkf.com/"
+      }
+    },
+    {
+      "name": "Response code: 199. This test case is not possible to implement for some platforms as this considered an invalid status code. Fallback into 200 test case",
```
rghetia: not sure what this means. spanStatus should be "OK' if it cannot be implemented and "UNKNOWN" if it can be implemented?

trace/HTTP.md
```diff
@@ -66,7 +66,20 @@ status code.
...
 Notes: 401 Unauthorized actually means unauthenticated according to RFC 7235, 3.1.
 
-The Status message should be the Reason-Phrase (RFC 2616 6.1.1) from the response status line (if available).
+The Status message should be the Reason-Phrase (RFC 2616 6.1.1) from the
+response status line (if available).
+
+### Client errors for client HTTP calls
+
+There are a number of client errors when trying to access http endpoint.
+Here are example of mapping those to the Open Census status codes.
```
rghetia: nit:s/example/examples/

trace/HTTP.md
```diff
@@ -117,3 +132,49 @@ known attributes/labels on supported tracing backends.
...
+Test cases for outgoing http calls are in the file
+[http-out-test-cases.json](http-out-test-cases.json).
+
+File consists of a set of test cases. Each test case represents outgoing http
+call, response it receives and the resulting span properties. It looks like
+this:
+
+``` json
+{
+"name": "Name is populated as a path",
```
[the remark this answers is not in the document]
rghetia: I agree. path should be avoided.

trace/HTTP.md
```diff
@@ -66,7 +66,20 @@ status code.
...
-The Status message should be the Reason-Phrase (RFC 2616 6.1.1) from the response status line (if available).
+The Status message should be the Reason-Phrase (RFC 2616 6.1.1) from the
+response status line (if available).
+
+### Client errors for client HTTP calls
+
+There are a number of client errors when trying to access http endpoint.
+Here are example of mapping those to the Open Census status codes.
+
+| Client error                 | Trace status code     |
```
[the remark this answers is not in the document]
SergeyKanzhelev: these are client errors when client wasn't able to reach server at all or was terminated before recieved a response

trace/HTTP.md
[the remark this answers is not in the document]
SergeyKanzhelev: This is an example of outgoing call. For outgoing call this spec calls for collecting a name as a path. Thus the test case

> Span name is formatted as:
> * /$path for outgoing requests.
> * /($path|$route) for incoming requests.

trace/http-out-test-cases.json
SergeyKanzhelev: on some platforms this test case is impossible to implement. I kept it as an entry, but actual parameters are for `200` test case

SergeyKanzhelev:
@ramonza @bogdandrutu sorry for delay. I addressed feedback.

trace/HTTP.md
```diff
@@ -66,7 +66,20 @@ status code.
...
 Notes: 401 Unauthorized actually means unauthenticated according to RFC 7235, 3.1.
 
-The Status message should be the Reason-Phrase (RFC 2616 6.1.1) from the response status line (if available).
+The Status message should be the Reason-Phrase (RFC 2616 6.1.1) from the
+response status line (if available).
+
+### Client errors for client HTTP calls
+
+There are a number of client errors when trying to access http endpoint. Here
+are examples of mapping those to the Open Census status codes.
```
ramonza:
```suggestion
are examples of mapping those to the OpenCensus status codes.
```

trace/HTTP.md
```diff
@@ -66,7 +66,20 @@ status code.
...
+
+| Client error                 | Trace status code     |
+|------------------------------|-----------------------|
+| DNS resolution failed        | 2 (UNKNOWN)           |
+| Request cancelled by caller  | 1 (CANCELLED)         |
+| Url cannot be parsed         | 3 (INVALID_ARGUMENT)  |
```
ramonza:
```suggestion
| URL cannot be parsed         | 3 (INVALID_ARGUMENT)  |
```

trace/HTTP.md
```diff
@@ -90,20 +103,22 @@ Implementations SHOULD create message event when response size is determined.
...
+
+| Attribute name            | Description                 | Type   |Example value              |
+|---------------------------|-----------------------------|--------|---------------------------|
+| "http.host"               | Request URL host            | string | `example.com:779`         |
+| "http.method"             | Request URL method          | string | `GET`                     |
+| "http.path"               | Request URL path. If empty - set to `/` | `/users/25f4c31d`  |
+| "http.route"              | Matched request URL route   | string | `/users/:userID`          |
+| "http.user_agent"         | Request user-agent. Do not inject attribute if user-agent is empty. | string | `HTTPClient/1.2` |
+| "http.status_code"        | Response status code        | int64  | `200`                     |
+| "http.url"                | Absolute request URL        | string | `https://example.com:779/path/12314/?q=ddds#123` |
```
ramonza: Should we specify how this URL is to be constructed from the request (host header, protocol, etc.)?

ramonza:
@SergeyKanzhelev the email address you use as `author` of the commit needs to be the same as the one you used to sign the CLA. I don't think I can override this (at least I have never done this before).

SergeyKanzhelev:
@ramonza I've signed with the Google account that has totally different e-mail address that I ever used on GitHub. I set GitHub account name on CLA page and just double checked it:

image

Can you please advice what would be my next steps?

ramonza:
@bogdandrutu can you help out here?

SergeyKanzhelev:
CLA is good for me in [this](https://github.com/census-instrumentation/opencensus-specs/pull/208) PR:

image

PR #184, opened 2018-09-24
Include status in client views

ramonza:


stats/HTTP.md
```diff
@@ -22,7 +22,7 @@ Buckets for distributions in default views are as follows:
 
 ### Measures
 
-Client stats are recorded for each individual HTTP request, including for each individual redirect (followed or not).
+Client stats are recorded for each individual HTTP request, including for each individual redirect (followed or not). All stats are recorded after request processing (usually after the response body has been fully read).
```
SergeyKanzhelev: comment on inclusion/exclusion of body read callbacks into the duration would be helpful. Since you are on it =)

ramonza: What is a body read callback?

SergeyKanzhelev: I meant do you include time on parsing of http body or not

ramonza: It depends on whether such processing happens before the body has been fully read. For example, in a streaming system you would process input as it is being read and then you would include this processing in the stats. But if you read it all into a temporary buffer and then process it then you would not include this time.

I think this is already entailed by the statement "after the response body has been fully read" so I don't want to add anything about *how* the response body is processed as this is likely to be language-specific.

PR #185, opened 2018-09-25
Make TraceContext the default propagation format

ramonza:


SergeyKanzhelev:
@bogdandrutu thanks for waiting. Let's see if it can be the only default by the first version release.

bogdandrutu:
I think we need to provide a way to make the transition for languages that already used b3 as default. @rghetia can you provide these instructions?

rghetia:
Transitioning Plan from B3 to TraceContext as default propagation.

**Release1**
By default read B3 andTraceContext format and write B3.
Users can explicitly use TraceContext format to write as they transition.

**Release2**
By default read B3 and TraceContext format and write TraceContext.
User can choose to optionally disable reading of B3 format. This can be useful to find out if there are still sources of B3 format that needs to transition.

**Release3**
Read and Write TraceContext by default.

rghetia:
Revised Text

### Transitioning Plan from B3 to TraceContext as default propagation.

#### Release X
Create a new propagator which can read/write B3 and TraceContext format. For lack of a better word , let's call it combined-propagator. By default it MUST read either B3 or TraceContext format and write ONLY B3 format. If both formats are present then read ONLY B3 format first. However, If there is a failure reading B3 format due to any reason then it MUST read TraceContext format if it is available.

Use combined-propagator as default propagator.

#### Release Y
Use the same combined-propagator as in Release X but with a different default behavior. By default it MUST read either TraceContext or B3 format and write ONLY TraceContext format. If both formats are present then it MUST read TraceContext format ONLY. However, If there is a failure in reading TraceContext format due to any reason then it MUST read B3 format if available.

Use combined-propagator as default propagator.

#### Release Z

Use TraceContext as a default propagator.

ramonza:
Why not have Release X read and write both? The transition from X -> Y is a silent change that could break existing users. I would prefer that this was an opt-in, rather than we just change the default underneath users. Changing to read and write both has the advantage that it doesn't break anyone but also allows users to gradually opt-in to TraceContext-only.

rghetia:
If it is opt-in then it is not a default.  

Whether we do read and write both or  write only one, the break will happen in both case. Break will happen only when one or more service is not upgraded at all (ie. they use pre-release X). 

The positive of both write is that it only takes two release cycle to change the default
Pre-X: IN(B3) -> OUT(B3)
X: IN(B3, TC) -> OUT(B3,TC)
Y: IN(TC)->OUT(TC)

The Positive for one write is that there is no duplication but it will take three release cycle to change the default.
Pre-X: IN(B3) -> OUT(B3)
X: IN(B3, TC) -> OUT(B3)
Y: IN(B3, TC)->OUT(TC)
Z: IN(TC) -> OUT(TC)

Issue #195, opened 2018-10-08
Decide on the specification for key names for log correlation


sebright:
This issue is for continuing the discussion about key names used in log correlation from https://github.com/census-instrumentation/opencensus-specs/pull/181#discussion_r219298220 .

/cc @rakyll @ramonza @adriancole @bogdandrutu

adriancole:
maybe add to the description the current state and the proposed one?

sebright:
I was planning to add the current state after I merge #181, since the key specification could still change.  I didn't want to forget to open an issue, though.

sebright:
There were two main suggested changes in https://github.com/census-instrumentation/opencensus-specs/pull/181#discussion_r219298220:

- Use shorter key names.
- Allow each implementation to use a format that is consistent with the language/logging framework that it supports.

sebright:
I thought that it might be better to add "opencensus" to avoid conflicting with key-value pairs that other libraries insert into the log entry or logging context.  However, conflicts probably won't be an issue if all log correlation implementations also provide a way to override the key names when necessary.

adriancole:
I think multiple tracing libraries are actually quite likely. Some tools
have tracing at very low levels (even pre 1.0 stuff). However, I do agree
that intuitive is nice regardless of this...

I also prefer traceId, spanId actually under the same logic you mention.

for why parentId? there still exist log based solutions which parse spans.
one example is cloud foundry, but there were others prior to that too.

personally I would like parentId to be opt-in ex no overhead unless
requested.

sebright:
I made a PR for shorter key names: #199

sebright:
I merged #199, so the current key names are "traceId", "spanId", and "traceSampled".

adriancole:
in general do we have a gate for removing draft? like a few libraries supporting it?

sebright:
Step 3 in https://github.com/census-instrumentation/opencensus-specs/blob/master/Process.md#step-3-specs-pull-request recommends creating a pull request implementing the change in at least one language, though I don't see any guidelines on testing the feature in a released library.

/cc @bogdandrutu

adriancole:
hmmm proposing a single pull request seems a really low barrier though
likely you could guess my opinion.

c24t:
> For the moment we have only Java and Go here. Can we get some input from other languages?

[...]

PR #199, opened 2018-10-11
Use shorter key names for log correlation (issue #195).

sebright:
This commit removes the "openCensus" prefix from all key names in the log
correlation spec.  The shorter key names are simpler and could improve
performance when the keys are included in the logs.

/cc @yurishkuro @rakyll @ramonza @SergeyKanzhelev

### DIFF
--- a/trace/LogCorrelation.md
+++ b/trace/LogCorrelation.md
@@ -54,5 +54,5 @@ and format the sampling decision as "true" or "false".
 
 Some logging frameworks allow the insertion of arbitrary key-value pairs into log entries.  When
 a log correlation implementation inserts tracing data by that method, the key names should be
-"opencensusTraceId", "opencensusSpanId", and "opencensusTraceSampled" by default.  The log
-correlation implementation may allow the user to override the tracing data key names.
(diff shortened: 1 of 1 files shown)


trace/LogCorrelation.md
```diff
@@ -54,5 +54,5 @@ and format the sampling decision as "true" or "false".
 
 Some logging frameworks allow the insertion of arbitrary key-value pairs into log entries.  When
 a log correlation implementation inserts tracing data by that method, the key names should be
-"opencensusTraceId", "opencensusSpanId", and "opencensusTraceSampled" by default.  The log
-correlation implementation may allow the user to override the tracing data key names.
+"traceId", "spanId", and "traceSampled" by default.  The log correlation implementation may allow
```
adriancole: probably a good name for traceSampled.. I've seen "sampled" but no strong feelings.

trace/LogCorrelation.md
```diff
@@ -54,5 +54,5 @@ and format the sampling decision as "true" or "false".
...
+the user to override the tracing data key names.
```
g-easy: Optional: "override the key names"

sebright: Done

PR #200, opened 2018-10-15
[WIP] Initial Draft for TagMap and its propagation.

rghetia:
Creating pull request to get initial feedback, specifically on the scope of a Tag.


tags/TagMapContext.md
```diff
@@ -0,0 +1,144 @@
...
+with a specific operation, such as an HTTP request. These `Tag`s are used to
+aggregate measurements in a [`View`](https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/DataAggregation.md#view) 
+according to unique value of the `Tag`s. The `Tag`s can also be used to filter (include/exclude)
+measurements in a `View`.
+
+# TagMap 
+`TagMap` is an abstract data type that represents collection of tags. 
+i.e., each key is associated with exactly one value, but multiple keys can be
+associated with the same value.  `TagMap` is serializable, and it represents
+all of the information that could be propagated across process boundaries.
```
bogdandrutu: inside the process and across process boundaries?

tags/TagMapContext.md
```diff
@@ -0,0 +1,144 @@
+# Summary
+A `Tag` is a key-value pair that can be used to label anything that is associated
+with a specific operation, such as an HTTP request. These `Tag`s are used to
+aggregate measurements in a [`View`](https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/DataAggregation.md#view) 
+according to unique value of the `Tag`s. The `Tag`s can also be used to filter (include/exclude)
+measurements in a `View`.
+
+# TagMap
```
bogdandrutu: Mention that this is the recommended name but languages can have a more language specific name like in go "tag.Map" because how the packages work in go.

tags/TagMapContext.md
```diff
@@ -0,0 +1,144 @@
...
+all of the information that could be propagated across process boundaries.
+
+## TagKey
+
+A string or string wrapper, that contains two parts, scope and key. The scope
+of a tag is used to determine if a tag should be propagated or not.
+
+
+**TagKey Syntax:**  
+Scope and key are separated by ':'.
```
bogdandrutu: where? I am confused about this "syntax" part. I understand that you have a Scope and a "Key" (string).

tags/TagMapContext.md
```diff
@@ -0,0 +1,144 @@
...
+|------|-------|
+| 0    | Process |
+| 1    | Internal-Service|
+| 2    | External-Service|
+
+
+##### Process
+A `TagKey` with the scope type *Process* is relevant only to stats collected by the process.
+Hence it should not be propagated across different processes. This is the default scope. If no
+scope is present in the `Tagkey` then it is treated as of Process scope type.
```
bogdandrutu: s/Tagkey/TagKey

tags/TagMapContext.md
```diff
@@ -0,0 +1,144 @@
...
+
+##### Internal-Service
```
bogdandrutu: You kind of make service a first class concept in OC by adding this. I see the value but I think we should allow users to achieve this in a different way (control the incoming/outgoing tags in the http/grpc plugins).

tags/TagMapContext.md
```diff
@@ -0,0 +1,144 @@
...
+Key is a string, with following restrictions:
+
+- Must contain only printable ASCII (codes between 32 and 126 inclusive with below Exception)
+- - Exceptions ':', ','
+- Must have length greater than zero and less than 256 combined with the scope.
+- Must not be empty.
+
+### Scope
+
+#### Scope Syntax
```
bogdandrutu: Same. Syntax for me refers to wire representation or string representation. Is this what you suggest here?

tags/TagMapContext.md
```diff
@@ -0,0 +1,144 @@
...
+A `TagKey` with the scope type *Internal-Service* is relevant to stats collected within an
+internal service which could be a single process or collection of process fulfilling the service.
+An internal services is the one that is not accessible from the public domain.
+An Example of internal service is a distributed caching service.
+
+A `TagKey` with this scope is propagated between different processes that makes up the internal
+service.
+
+
+##### External-Service
```
bogdandrutu: Same problem. I don't think OC knows about "Service"

bogdandrutu:
Overall I think we mix the definition of the concepts with wire format. We should focus on what a TagKey is: scope + name (key) for example and not how it is formatted.

How we encode these data scope + name is a different problem.

tags/TagMapContext.md
bogdandrutu: I think the library should be aware of the Scope to do the correct thing in terms of propagating or not these tags.

adriancole:
once upon a time this was called TagSet in goog right? I think I saw this
couple years ago in another cpp goog project maybe?

tags/TagMapContext.md
```diff
@@ -0,0 +1,144 @@
...
+# TagMap 
+`TagMap` is an abstract data type that represents collection of tags. 
+i.e., each key is associated with exactly one value, but multiple keys can be
+associated with the same value.  `TagMap` is serializable, and it represents
+all of the information that could be propagated across process boundaries.
+
+## TagKey
+
+A string or string wrapper, that contains two parts, scope and key. The scope
+of a tag is used to determine if a tag should be propagated or not.
```
adriancole: scope is the extent of propagation right? it wouldn't make sense to never propagate even internal to process right?

tags/TagMapContext.md
```diff
@@ -0,0 +1,144 @@
...
+
+**TagKey Syntax:**  
+Scope and key are separated by ':'.
+```
+<Scope>:<Key>
+```
+
+**Scope**
+Scope of a `TagKey` defines a boundary within which the `TagKey` is relevant. For example 
+a `TagKey` http_route on frontend service could have a scope that spans across multiple services
```
adriancole: this is interesting. http route is not something I thought of as propagating that far. maybe something like endpoint (edge request)?

tags/TagMapContext.md
```diff
@@ -0,0 +1,144 @@
...
+that contributes to fulfilling the http request. On the other hand if you are measuring latency
+of key individual functions in a ProductCatalog service then the scope of a `TagKey` func_name
+is most likely limited to a process in ProductCatalog service.
+
+The scope is used to determine whether a Tag should be propagated or not.
```
adriancole: need to introduce scope types earlier as it gets weird with statements that feel boolean such as this

tags/TagMapContext.md
```diff
@@ -0,0 +1,144 @@
...
+- Scope MUST always be in a Type-Value pair.
+
+There are following scopes types.
+
+#### Scope Types
+Scope Type is represented using code to reduce the size.
+
+| Code | Type  |
+|------|-------|
+| 0    | Process |
```
adriancole: add a column and define these in terms of where they stop and/or if this is an exported scope (serialized in headers).  also add a sentence hinting they are elaborated below

tags/TagMapContext.md
adriancole: people have re-entrant processes and also multiple tracers. I might clarify that this scope is per tracer instance

tags/TagMapContext.md
adriancole: a lot not defined.. ex how we even can tell internal from external.

what about starting with basics? remote and not remote?

tags/TagMapContext.md
```diff
@@ -0,0 +1,144 @@
...
+TagMap should be encoded using [BinaryEncoding](https://github.com/census-instrumentation/opencensus-specs/tree/master/encodings)
+and propagated using gRPC metadata opencensus-tag-bin.
+
+### Over HTTP
+
+**TBD**
+                                                
+## Error handling
+
+- There are no partial failures.  The result of encoding or decoding
```
adriancole: error handling should not break callers. this section makes me think it is ok to raise error

tags/TagMapContext.md
rghetia: right and right. What I meant was at every propagation point scope is used to determine whether to propagate or not. But I will make it clear to not sound like it is a binary property (propagate or not-propagate) of a tag.

yurishkuro:
In several places the wording couples tags with stats/measurements, as though it is is the only usage of tags. Is that intentional?  What if someone wants to define a process-level tag that is used in logs, but not metrics?

tags/TagMapContext.md
rghetia: ok.

tags/TagMapContext.md
rghetia: What would be a typical value of tagkey edge-request?

tags/TagMapContext.md
yurishkuro: This one is very confusing to me. First, the description is not particularly clear about what "internal service" means. But more importantly, how can the instrumentation library know about that? E.g. say I am making a request to some host:port, how can OC tell that it's a request to "internal service"?

tags/TagMapContext.md
rghetia: Comment#1 - Per tracer instance
I will clarify it.

Comment#2  - remote and not remote
Good point. Let me make the re-draft the spec.

Comment#3 - Error handling.
Let me separate out the encoding and error handling separately.
The intention here is not to break the call if there is error decoding the TagMap. But rather to accept or not accept partial data if there is an error. I'll clarify it.

rghetia:
@yurishkuro There is no reason these tags cannot be used for logs or trace. I'll make the change appropriately.

tags/TagMapContext.md
rghetia: I have re-drafted the spec based on suggestion and removed the concept of internal/external service. I will post the new draft soon.

tags/TagMapContext.md
```diff
@@ -0,0 +1,119 @@
...
+A receiver at the edge of an entity converts source of all incoming tags to remote tags. Once the
+source of a tag is designated as remote it stays as remote for the rest of the request processing.
+This is applicable to reentrant processes as well ie. A tag T is created in an entity E and goes
+through process P as local tag. It then leaves entity E and 
+returns back and goes through Process P again. Tag T is considered as remote tag by process P
+second time. (This only matters if Process P differentiates between remote and local tags. For e.g
+it ignores all remote tags and uses only local tags for aggregation. Note )
+
+
+A tag propagator should provide an option to declare that it is running on the edge of an entity.
```
yurishkuro: what is the use case that cares about "source"?

rghetia: Some process or micro-service could ignore remote tag for their metrics or logging or tracing but it would still forward the tags.

tags/TagMapContext.md
```diff
@@ -0,0 +1,119 @@
...
+## Wire Format
+TBD
+
+### Over gRPC
+TagMap should be encoded using [BinaryEncoding](https://github.com/census-instrumentation/opencensus-specs/tree/master/encodings)
+and propagated using gRPC metadata opencensus-tag-bin.
+
+### Over HTTP
+
+**TBD**
```
SergeyKanzhelev: https://github.com/w3c/correlation-context/blob/master/correlation_context/HTTP_HEADER_FORMAT.md ?

tags/TagMapContext.md
```diff
@@ -0,0 +1,119 @@
...
+
+## TagKey
+
+`TagKey` consists of source, metadata and name.
+
+### Source
+
+Source of a tag is either remote or local.
+ 
+**Local Source**
```
SergeyKanzhelev: who decides whether local tags should be propagated? Should there be "local-only" tags?

tags/TagMapContext.md
yurishkuro: That use case does not require pulling the "source" as a top-level concept, nor does the "source" address that use case fully. For example, I may want to set some tags in the application and propagate them down, yet I don't want them to be used as dimensions in the metrics. For example, I may want to propagate the UUID of the top-level user through the whole call graph, yet I'd never want to partition metrics by that value.

rghetia: I had a chat with @bogdandrutu and he had suggested view definition already takes care of whether to use the tag for metric partition or not. So it seems that all we need is
1. A scope to indicate that the tag is of local significance (process level) and should not be propagated.
2. Exclude/Include criteria for incoming and outgoing tags.

tags/TagMapContext.md
rghetia: From an initial glance, it seems it would work.

tags/TagMapContext.md
```diff
@@ -0,0 +1,119 @@
...
+it ignores all remote tags and uses only local tags for aggregation. Note )
+
+
+A tag propagator should provide an option to declare that it is running on the edge of an entity.
+
+### Metadata
+
+Metadata of a `TagKey` is used to provide additional information about the `TagKey`. The 
+metadata can be used to filter tags. Tags are filtered on the edge of an entity. Metadata is 
+simply a collection of an arbitraty string.
```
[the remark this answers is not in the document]
rghetia: I am going to remove metadata. The naming of the tag key should provide sufficient information and should not require any additional metadata. I had misunderstood the requirement when I initially spoke to @bogdandrutu .

tags/TagSet.md
```diff
@@ -0,0 +1,109 @@
...
+**Local Scope**
+Tag with `Local` scope are used within the process it created. Such tags are not propagated
+across process boundaries. Even if the process is reentrant the tag MUST be excluded from
+propagation when the call leaves the process.
+
+**Non-Local Scope**
+If a tag is created with the `Non-Local` scope then it is propagated across process boundaries subject 
+to outgoing and incoming (on remote side) filter criteria. See `TagFilter` in [Tag Propagation](#Tag Propagation).
+
+# TagSet
```
bogdandrutu: Should this be TagMap as we discussed? Same for the file name.

tags/TagSet.md
```diff
@@ -0,0 +1,109 @@
...
+A `Tag` consists of TagScope, TagKey, and TagValue.
+
+## TagKey
+
+`TagKey` is the name of the Tag. TagKey along with `TagValue` is used to aggregate
+and group stats, annotate traces and logs.
+
+**Restrictions**
+- Must contain only printable ASCII (codes between 32 and 126 inclusive with below Exception)
+- - Exceptions '=', ','
```
bogdandrutu: Where is this coming from?

tags/TagSet.md
```diff
@@ -0,0 +1,109 @@
...
+- Must have length greater than zero and less than 256 combined with the source and metadata.
```
bogdandrutu: What metadata? What source?

tags/TagSet.md
```diff
@@ -0,0 +1,109 @@
...
+- Must not be empty.
+
+## TagValue
```
bogdandrutu: Say that can be empty

tags/TagSet.md
```diff
@@ -0,0 +1,109 @@
...
+
+`TagValue` is a string or string wrapper. It MUST contain only printable ASCII (codes between
+32 and 126)
+
+## TagScope
+
+`TagScope` is used to determine the scope of a `Tag`. For now, it has two possible values,
+Local or Non-Local. In future, additional values be added to address specific situations.
```
bogdandrutu: Maybe local and global?

tags/TagSet.md
```diff
@@ -0,0 +1,109 @@
...
+different then that for forwarding.
+
+If no filter is specified for receiving then all `Tag`s are received. 
+If no filter is specified for forwarding then all `Tag`s are forwarded except those that have `Local Scope`.
+
+### TagFilter
+Tag Filter consists of `TagFilterType` and `TagFilterRegexp`.
+
+#### TagFilterType
+There are three types of Filter
```
bogdandrutu: I can see only 2.

adriancole:
local and remote seems more immediately understandable vs other choices said

adriancole:
not sure how many saw this, but back in january, we collected feedback about this in brave. Should be helpful in design though a lot of this looks compatible https://github.com/openzipkin/brave/pull/577

tags/TagSet.md
yurishkuro: `Scope.(Process | Local)`, `Scope.(Request | Transaction | Execution)`.

`global` sounds a bit scary & imprecise. `remote` hints at origin rather than scope.

tags/TagSet.md
rghetia: earlier one of the comment was to use TagSet. I can change it back to TagMap.

tags/TagSet.md
```diff
@@ -0,0 +1,109 @@
...
+# TagSet 
+`TagSet` is an abstract data type that represents collection of tags. 
+i.e., each key is associated with exactly one value, but multiple keys can be
+associated with the same value.  `TagSet` is serializable, and it represents
+all of the information that could be propagated inside the process and across process boundaries.  
+`TagSet` is a recommended name but languages can have more language specific name.
+
+## Limits
+Combined size of all `Tag`s should not exceed 8192 bytes before encoding.
+The size restriction applies to the deserialized tags so that the set of decoded
```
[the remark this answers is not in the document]
rghetia: make sense. I'll remove it.

tags/TagSet.md
```diff
@@ -0,0 +1,109 @@
...
+ `TagSet`s is independent of the encoding format.
+
+## TagSet Propagation
+`TagSet` may be propagated across process boundaries or across any arbitrary boundaries for various
```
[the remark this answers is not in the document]
rghetia: I'll change it to "project-id".

tags/TagSet.md
rghetia: Earlier I had filter type None which I deemed not necessary. I'll fix it.

tags/TagSet.md
```diff
@@ -0,0 +1,109 @@
...
+If no filter is specified for forwarding then all `Tag`s are forwarded except those that have `Local Scope`.
+
+### TagFilter
+Tag Filter consists of `TagFilterType` and `TagFilterRegexp`.
+
+#### TagFilterType
+There are three types of Filter
+
+**Exclude**
+If the `TagFilterType` is Exclude then any `Tag`s whose `TagKey` matches the `TagFilterRegexp`
```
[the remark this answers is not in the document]
rghetia: may be add MatchOperator. Start with support for equals and LIKE. If need be it can be expanded to other operators.

tags/TagSet.md
```diff
@@ -0,0 +1,109 @@
...
+MUST be excluded. Unmatched `Tag`s are subjected to further processing against remaining `TagFilter`s.
+
+**Include**
+If the `TagFilterType` is Include then any `Tag`s whose `TagKey` matches the `TagFilterRegexp`
+MUST be included. Unmatched `Tag`s are subjected to further processing against remaining `TagFilter`s.
```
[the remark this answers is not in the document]
rghetia: agreed. Let me add the concept of FilterList as an ordered list of TagFilter.

tags/TagSet.md
```diff
@@ -0,0 +1,109 @@
...
+and propagated using gRPC metadata opencensus-tag-bin.
+TBD: how and if `TagScope` should be propagated.
+
+#### Over HTTP
+
+TBD: W3C [correlation context](https://github.com/w3c/correlation-context/blob/master/correlation_context/HTTP_HEADER_FORMAT.md)
+may be an appropriate choice.
+
+
+TBD: how and if `TagScope` should be propagated.
```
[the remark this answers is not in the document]
rghetia: I don't have any use case in mind. I added the statement to see if others know of any use case where it needs propagation.

tags/TagSet.md
yurishkuro: | is "or", so values:
* local: Process or Local
* global: Request or Transaction or Execution

rghetia: so you are suggesting ScopeType (local | gloabl) and ScopeValues (according to Scope type). Since scope is not propagated how will these values be used? Will they be used by backend?

yurishkuro: I was just suggesting alternative names for local/global

adriancole:
I think scope should be limited to propagation. global makes no sense.
remote does. you need more different people commenting here

tags/TagSet.md
bogdandrutu: I agree global is not that good. Probably Request or Transaction are better names.

tags/TagMap.md
```diff
@@ -0,0 +1,122 @@
...
+**Include**
+If the `TagFilterAction` is Include then any `Tag` whose `TagKey` evaluates to true 
+with the condition (`TagFitlerMatchOperator` and `TagFilterMatchString`)
+MUST be included. 
+  
+#### TagFilterMatchOperator
+
+| Operator | Description |
+|----------|-------------|
+| EQUAL | The condition is evaluated to true if `TagKey` is exactly same as `TagFilterMatchString` |
```
bogdandrutu: Maybe we want also HAS_PREFIX (very useful in general in the internal code).

tags/TagMap.md
```diff
@@ -0,0 +1,122 @@
...
+reasons. For example, one may propagate 'project-id' Tag across all micro-services to break down metrics
+by 'project-id'. Not all `Tag`s in a `TagMap` should be propagated and not all `Tag`s in a `TagMap`
+should be accepted from a remote peer. Hence, `TagMap` propagator must allow specifying an optional
+list of ordered `TagFilter`s for receiving `Tag`s or for forwarding `Tag`s or for both. A `TagFilter` list 
+for receiving MAY be different then that for forwarding.
+
+If no filter is specified for receiving then all `Tag`s are received. 
+If no filter is specified for forwarding then all `Tag`s are forwarded except those that have `Local Scope`.
+
+### TagFilter
```
bogdandrutu: We've seen some use-cases where filters are good for Views as well. Build distribution X (with tags x,y,z) where tagKey t has value foo. So that may be a different type of TagFilter, but want to say about that use-case because that may influence the name of this field. Maybe consider TagPropagationFilter?

tags/TagMap.md
```diff
@@ -0,0 +1,124 @@
...
+### TagPropagationFilter
+Tag Propagation Filter consists of action (`TagFilterAction`) and condition (`TagFilterMatchOperator` and
+`TagFilterMatchString`). A `TagKey` 
+is evaluated against condition of each `TagPropagationFilter` in order. If the condition is evaluated
+to true then action is taken according to `TagFilterAction` and filter processing is stopped.
+If the condition is evaluated to false then the `TagKey` is processed against next `TagPropagationFilter`
+in the ordered list. If none of the condition is evaluated to true then the default
+action is **Exclude**.
+
+#### TagFilterAction
```
bogdandrutu: TagPropagationFilter.Action? same for all classes?

rghetia: done.

tags/TagMap.md
```diff
@@ -0,0 +1,125 @@
...
+is evaluated against condition of each `TagPropagationFilter` in order. If the condition is evaluated
+to true then action is taken according to `TagPropagationFilterAction` and filter processing is stopped.
+If the condition is evaluated to false then the `TagKey` is processed against next `TagPropagationFilter`
+in the ordered list. If none of the condition is evaluated to true then the default
+action is **Exclude**.
+
+#### TagPropagationFilterAction
+There are two actions to apply to a Tag.
+
+**Exclude**
```
bogdandrutu: I would make this an interface that way we can add for example sanitization or mutation of the TagValue. Include exclude is fine to be the first supported actions.

PR #201, opened 2018-10-16
Implementations should provide a way to avoid undesirable propagation

ramonza:
Sometimes it is not desirable to propagate tracing metadata at all.
For example, if you're calling an external API.

### DIFF
--- a/trace/HTTP.md
+++ b/trace/HTTP.md
@@ -4,7 +4,7 @@ This document explains tracing of HTTP requests with OpenCensus.
 
 ## Spans
 
-Implementations MUST create a span for outgoing requests at the client and a span for incoming 
+Implementations MUST create a span for outgoing requests at the client and a span for incoming
 requests at the server.
 
 Span name is formatted as:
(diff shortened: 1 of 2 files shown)


trace/HTTP.md
```diff
@@ -29,24 +29,28 @@ incoming requests should be a span kind of SERVER.
 
 Propagation is how SpanContext is transmitted on the wire in an HTTP request.
 
-Implementations MUST allow users to set their own propagation format and MUST provide an 
+Implementations MUST allow users to set their own propagation format and MUST provide an
 implementation for B3 at least.
 
+Implementations should provide a way to avoid propagating SpanContext altogether from both
```
bogdandrutu: Not sure what the inbound requests propagation means. Do you mean trusting the incoming SpanContext?

trace/HTTP.md
```diff
@@ -29,24 +29,28 @@ incoming requests should be a span kind of SERVER.
 
 Propagation is how SpanContext is transmitted on the wire in an HTTP request.
 
-Implementations MUST allow users to set their own propagation format and MUST provide an 
+Implementations MUST allow users to set their own propagation format and MUST provide an
 implementation for B3 at least.
```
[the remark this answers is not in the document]
adriancole: should be in a different PR if action taken as it would be surprising to see this as a side effect of the work titled "Implementations should provide a way to avoid undesirable propagation"

trace/gRPC.md
```diff
@@ -25,13 +25,17 @@ For backends that does not support Span.Kind, the exported span names can be pre
 
 Propagation is how SpanContext is transmitted on the wire in an gRPC request.
 
-The propagation MUST inject a SpanContext (as a gRPC metadata `grpc-trace-bin`) and MUST extract 
+The propagation MUST inject a SpanContext (as a gRPC metadata `grpc-trace-bin`) and MUST extract
 a SpanContext from the gRPC metadata. The serialization format is defined
 [here](../encodings/BinaryEncoding.md).
 
+Implementations should provide a way to avoid propagating SpanContext altogether from both
```
adriancole: is there a way to avoid copy/pasta?

trace/HTTP.md
```diff
@@ -29,24 +29,28 @@ incoming requests should be a span kind of SERVER.
...
 
+Implementations should provide a way to avoid propagating SpanContext altogether from both
+inbound requests and on outbound requests while still generating spans as usual
+around the inbound/outbound request.
```
adriancole: This section should have a use case as this implies complexity. For example, you'd need some sort of functionality to express which endpoints are those which you'd not want to propagate downstream to. Also we are moving features here which could be addressed individually with middleware (header scrubbing for various reasons can be site specific)

ramonza:
Thanks for the review @adriancole. I agree that it should have been discussed more and I will do that in an issue in future before sending a PR. Here is an issue that I filed for this: https://github.com/census-instrumentation/opencensus-specs/issues/204

ramonza:
Thanks for the review @adriancole. I agree that it should have been discussed more and I will do that in an issue in future before sending a PR. Here is an issue that I filed for this: https://github.com/census-instrumentation/opencensus-specs/issues/204

songy23:
Superseded by `TagMetadata`.

Issue #202, opened 2018-10-17
Consider requiring a feature to be implemented in a few libraries before making the spec final


sebright:
See https://github.com/census-instrumentation/opencensus-specs/issues/195#issuecomment-430091302.

We could update https://github.com/census-instrumentation/opencensus-specs/blob/a6ad82f4ee21038c33d5282ebb4979d08808e6b5/Process.md#step-3-specs-pull-request to require each new feature to be merged with a "draft" label.  Then, after the feature has been implemented in a few libraries and has been shown to work well, we could remove the "draft" label.

/cc @adriancole @bogdandrutu

bogdandrutu:
+1 I think this is a good idea.

PR #203, opened 2018-10-17
Finalize the log correlation spec by removing "draft" from the title.

sebright:
This commit finalizes spec, since I think we have agreement on key names in #195 after they were shortened to "traceId", "spanId", and "traceSampled".

/cc @yurishkuro @rakyll @ramonza @SergeyKanzhelev @c24t

### DIFF
--- a/trace/LogCorrelation.md
+++ b/trace/LogCorrelation.md
@@ -1,7 +1,4 @@
-# Log Correlation (draft)
-
-This specification is not done until the key names below are finalized (issue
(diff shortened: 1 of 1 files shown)


sebright:
@bogdandrutu I opened an issue for deciding on the requirements for making a specification final: #202.  I wanted to remove "draft" from this feature because it has already been implemented in two libraries, and I think I resolved the issues from the PR where I added "draft" in #199.

sebright:
Thanks!

PR #206, opened 2018-11-02
Add measures that are updated for each message

ramonza:
For streaming RPCs, especially long-lived ones, we want to have
stats for each message as it is sent/received, rather than waiting
until the end of the RPC to record stats.

### DIFF
--- a/stats/gRPC.md
+++ b/stats/gRPC.md
@@ -29,19 +29,22 @@ Buckets for distributions in default views are as follows:
 
 Client stats are recorded at the end of each outbound RPC.
 
-| Measure name                             | Unit | Description                                                                                   |
-|------------------------------------------|------|-----------------------------------------------------------------------------------------------|
-| grpc.io/client/sent_messages_per_rpc     | 1    | Number of messages sent in the RPC (always 1 for non-streaming RPCs).                         |
-| grpc.io/client/sent_bytes_per_rpc        | By   | Total bytes sent across all request messages per RPC.                                         |
-| grpc.io/client/received_messages_per_rpc | 1    | Number of response messages received per RPC (always 1 for non-streaming RPCs).               |
-| grpc.io/client/received_bytes_per_rpc    | By   | Total bytes received across all response messages per RPC.                                    |
-| grpc.io/client/roundtrip_latency         | ms   | Time between first byte of request sent to last byte of response received, or terminal error. |
-| grpc.io/client/server_latency            | ms   | Propagated from the server and should have the same value as "grpc.io/server/latency".        |
-| grpc.io/client/started_rpcs              | 1    | The total number of client RPCs ever opened, including those that have not completed.         |
+| Measure name                              | Unit | Description                                                                                   |
+|-------------------------------------------|------|-----------------------------------------------------------------------------------------------|
+| grpc.io/client/sent_messages_per_rpc      | 1    | Number of messages sent in the RPC (always 1 for non-streaming RPCs).                         |
+| grpc.io/client/sent_bytes_per_rpc         | By   | Total bytes sent across all request messages per RPC.                                         |
+| grpc.io/client/received_messages_per_rpc  | 1    | Number of response messages received per RPC (always 1 for non-streaming RPCs).               |
+| grpc.io/client/received_bytes_per_rpc     | By   | Total bytes received across all response messages per RPC.                                    |
+| grpc.io/client/roundtrip_latency          | ms   | Time between first byte of request sent to last byte of response received, or terminal error. |
+| grpc.io/client/server_latency             | ms   | Propagated from the server and should have the same value as "grpc.io/server/latency".        |
+| grpc.io/client/started_rpcs               | 1    | The total number of client RPCs ever opened, including those that have not completed.         |
+| grpc.io/client/sent_bytes_per_message     | By   | Total bytes sent, recorded for each message in the RPC.                                       |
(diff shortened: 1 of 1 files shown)


stats/gRPC.md
```diff
@@ -66,6 +69,11 @@ The following set of views are considered minimum required to monitor client-sid
 | grpc.io/client/roundtrip_latency      | grpc.io/client/roundtrip_latency      | distribution | grpc_client_method                     |
 | grpc.io/client/completed_rpcs         | grpc.io/client/roundtrip_latency      | count        | grpc_client_method, grpc_client_status |
 | grpc.io/client/started_rpcs           | grpc.io/client/started_rpcs           | count        | grpc_client_method                     |
+| grpc.io/client/sent_message_bytes     | grpc.io/client/sent_message_bytes     | distribution | grpc_client_method                     |
```
zhangkun83: You'd probably want the names to be consistent with the existing metrics such as "sent_bytes_per_rpc". So this would be "sent_bytes" instead.

stats/gRPC.md
```diff
@@ -66,6 +69,11 @@ The following set of views are considered minimum required to monitor client-sid
...
+| grpc.io/client/received_message_bytes | grpc.io/client/received_message_bytes | distribution | grpc_client_method                     |
+| grpc.io/client/sent_message_count     | grpc.io/client/sent_message_bytes     | count        | grpc_client_method                     |
```
zhangkun83: Should be referred as "count", not "bytes".

songy23:
Friendly ping @ramonza

stats/gRPC.md
songy23: This has the same rational to https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/gRPC.md#why-are-completed_rpcs-views-defined-over-latency-measures. We can calculate the message count by how many times "sent_message_bytes" are recorded.

ramonza: yes, I will remove the _count views, since the count will be in the distribution of the _bytes views

stats/gRPC.md
ramonza: I wanted it to be clear that these are recorded *for each message* rather than for the whole RPC.

songy23:
Friendly ping @zhangkun83 and @bogdandrutu. 

Need this PR to unblock https://github.com/census-instrumentation/opencensus-java/pull/1563.

stats/gRPC.md
zhangkun83: "sent_message_bytes" doesn't make that more clear. To a random reader, the *message* here may be (mis)taken as describing the source of the bytes (message rather than metadata) instead of the granularity at which they are counted.

With the same idea as "sent_bytes_per_rpc", I would suggest "sent_bytes_per_message"

songy23: FYI the names are updated.

bogdandrutu:
@zhangkun83 do you mean that at one point you will get two numbers:
* num_messages -> not always 1
* size_for_all_these_messages

Is this the case?

zhangkun83:
I get notification about each message, while the size of the message is not always available. On the other hand, I get notification about a chunk of outbound or inbound bytes, which are not limited to a single message (could be partial message and could be across multiple messages). I do know when the RPC starts and finishes.

For the existing metrics, I get the message count per RPC from the first notification, and the byte count per RPC from the second notification.

songy23:
> GRPC library can try to get a best estimation, where message counts will be accurate, but the size distribution won't be 100% accurate.

I think this makes more sense. We can mention in the description of these metrics that when compression is used, the bytes histogram distribution may not be always correct. WDYT?

zhangkun83:
We can't guarantee the distribution is accurate even without stream compression, because there is no timing guarantee between the two types of notifications (message vs. byte blobs) in any case.

songy23:
I see, in this case the current metrics for the sent/received bytes only represent the distribution of bytes by each `record` call, and that's not correlated with the "per-message" distribution.

In this case I would recommend while keeping the current "per-record" distribution of bytes (IMO they're still useful), we also add *_MESSAGE_COUNT measures to get the accurate count.

zhangkun83:
It depends on how you define the "per-record" call.

If it's recorded per arbitrary chunk of data (not necessarily per message), having the distribution doesn't seem useful.

If you still define it as "per-message", GRPC would still call `record()` on a per-message basis, while the size of each message may not be accurate (their sum would be), and you wouldn't need the `*_MESSAGE_COUNT` measures because the number of the distribution data points still aligns with the number of messages.

songy23:
> If it's recorded per arbitrary chunk of data (not necessarily per message), having the distribution doesn't seem useful.

> If you still define it as "per-message", GRPC would still call record() on a per-message basis, while the size of each message may not be accurate (their sum would be),

Chatted offline, with the stream compression it will be very difficult to get the exact real-time per-message bytes. IMO it will be confusing to export a distribution of random chunks of bytes (looks like we had similar issue in Tracing before and we made the bytes optional). For now I'm leaning towards just having two accumulations of bytes and counts, since the cumulative stats are the only "exact" stats we can get.

WDYT? @bogdandrutu

bogdandrutu:
Is this true in all the languages or just a Java problem?

zhangkun83:
With stream-based compression it's impossible to accurately associate wire bytes with messages, and that's by design for all languages.

ramonza:
Closing this in favor of #212

PR #212, opened 2018-11-16
Add gRPC metrics that allow real-time reporting.

songy23:
Supersedes https://github.com/census-instrumentation/opencensus-specs/pull/206.

Use plain accumulations for streaming bytes and message counts. We won't be able to associate wire bytes with messages with stream-based compression.

### DIFF
--- a/stats/gRPC.md
+++ b/stats/gRPC.md
@@ -29,19 +29,22 @@ Buckets for distributions in default views are as follows:
 
 Client stats are recorded at the end of each outbound RPC.
 
-| Measure name                             | Unit | Description                                                                                   |
-|------------------------------------------|------|-----------------------------------------------------------------------------------------------|
-| grpc.io/client/sent_messages_per_rpc     | 1    | Number of messages sent in the RPC (always 1 for non-streaming RPCs).                         |
-| grpc.io/client/sent_bytes_per_rpc        | By   | Total bytes sent across all request messages per RPC.                                         |
-| grpc.io/client/received_messages_per_rpc | 1    | Number of response messages received per RPC (always 1 for non-streaming RPCs).               |
-| grpc.io/client/received_bytes_per_rpc    | By   | Total bytes received across all response messages per RPC.                                    |
-| grpc.io/client/roundtrip_latency         | ms   | Time between first byte of request sent to last byte of response received, or terminal error. |
-| grpc.io/client/server_latency            | ms   | Propagated from the server and should have the same value as "grpc.io/server/latency".        |
-| grpc.io/client/started_rpcs              | 1    | The total number of client RPCs ever opened, including those that have not completed.         |
+| Measure name                              | Unit | Description                                                                                   |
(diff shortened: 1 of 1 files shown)


stats/gRPC.md
```diff
@@ -29,27 +29,32 @@ Buckets for distributions in default views are as follows:
...
+| grpc.io/client/received_messages_per_rpc  | 1    | Number of response messages received per RPC (always 1 for non-streaming RPCs).               |
+| grpc.io/client/received_bytes_per_rpc     | By   | Total bytes received across all response messages per RPC.                                    |
+| grpc.io/client/roundtrip_latency          | ms   | Time between first byte of request sent to last byte of response received, or terminal error. |
+| grpc.io/client/server_latency             | ms   | Propagated from the server and should have the same value as "grpc.io/server/latency".        |
+| grpc.io/client/started_rpcs               | 1    | The total number of client RPCs ever opened, including those that have not completed.         |
+| grpc.io/client/sent_messages_per_method   | 1    | Total messages sent per method.                                    |
+| grpc.io/client/received_messages_per_method | 1  | Total messages received per method.                                |
+| grpc.io/client/sent_bytes_per_method      | By   | Total bytes sent per method, recorded for each message.           |
```
zhangkun83: Not necessarily recorded for each message. Maybe "recorded real-time as bytes are sent".

songy23: SG, updated.

songy23:
PTAL @bogdandrutu

stats/gRPC.md
```diff
@@ -59,13 +64,17 @@ https://github.com/grpc/grpc/blob/master/doc/statuscodes.md
...
+| View name                                 | Measure                                   | Aggregation  | Tags                                   |
+|-------------------------------------------|-------------------------------------------|--------------|----------------------------------------|
+| grpc.io/client/sent_bytes_per_rpc         | grpc.io/client/sent_bytes_per_rpc         | distribution | grpc_client_method                     |
+| grpc.io/client/received_bytes_per_rpc     | grpc.io/client/received_bytes_per_rpc     | distribution | grpc_client_method                     |
+| grpc.io/client/roundtrip_latency          | grpc.io/client/roundtrip_latency          | distribution | grpc_client_method                     |
+| grpc.io/client/completed_rpcs             | grpc.io/client/roundtrip_latency          | count        | grpc_client_method, grpc_client_status |
+| grpc.io/client/started_rpcs               | grpc.io/client/started_rpcs               | count        | grpc_client_method                     |
+| grpc.io/client/sent_messages_per_method      | grpc.io/client/sent_messages_per_method      | count | grpc_client_method                     |
```
bogdandrutu: I don't think this are necessary in the default views. I would put them in the extra views for the moment. We probably want to have these measurements not enabled by default to avoid unnecessary overhead for unary calls (probably the majority of the calls).

songy23: Yes I agree. Made them optional for the moment, and add an FAQ about these views.

songy23:
Thanks!

PR #215, opened 2018-11-21
Fix stats.proto broken link

simonz130:


johnchildren:
Also spotted this, for info the trace data model link is also broken, but probably best fixed in this PR.

PR #220, opened 2018-12-31
add instrumenting directory for docs on writing integrations

tsloughter:
I wasn't sure on the best directory layout for this was but I wanted to suggest moving documentation on instrumenting libraries for common protocols/interfaces like http, grpc, sql, etc into a common directory.

Figured I'd open a PR with one of the possibilities instead of opening an issue.


odeke-em:
Thanks for taking a stab at this @tsloughter! My personal opinion is that gRPC, Thrift and HTTP (etc) are transports hence are disjoint from integrations whereas writing SQL, Redis, Memcached are applications for which integrations can be written.

tsloughter:
@odeke-em true, but I don't think instrumentation of libraries that access applications should vary across language/implementation either.

What I've seen so far is while all grpc libs would use stats with a prefix of `grpc.io/client/` and `grpc.io/server/` the SQL libraries would use stats named like `go_sql_client` and `java_sql_client`. I'd argue it is better to define a similar shared set of measurement and view names (and suggested attributes on a span) whether it is a transport or a common application type.

bogdandrutu:
please rebase :)

tsloughter:
Rebased.

SergeyKanzhelev:
this file https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/http-out-test-cases.json should be moved alongside the http.md

PR #223, opened 2019-01-08
Fix trace data model link

c24t:
As per https://github.com/census-instrumentation/opencensus-specs/pull/215#issuecomment-441080424.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,5 +162,5 @@ The key elements the API MUST provide are:
 [RFC2119]: https://www.ietf.org/rfc/rfc2119.txt
 [StatsDataModel]: https://github.com/census-instrumentation/opencensus-proto/blob/master/src/opencensus/proto/stats/v1/stats.proto
 [DistributedTracingSpecs]: https://github.com/w3c/distributed-tracing
-[TraceDataModel]: https://github.com/census-instrumentation/opencensus-proto/blob/master/opencensus/proto/trace/v1/trace.proto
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -162,5 +162,5 @@ The key elements the API MUST provide are:
 [RFC2119]: https://www.ietf.org/rfc/rfc2119.txt
 [StatsDataModel]: https://github.com/census-instrumentation/opencensus-proto/blob/master/src/opencensus/proto/stats/v1/stats.proto
 [DistributedTracingSpecs]: https://github.com/w3c/distributed-tracing
-[TraceDataModel]: https://github.com/census-instrumentation/opencensus-proto/blob/master/opencensus/proto/trace/v1/trace.proto
+[TraceDataModel]: https://github.com/census-instrumentation/opencensus-proto/blob/master/src/opencensus/proto/trace/v1/trace.proto
```
mayurkale22: I think we should use permalink: https://github.com/census-instrumentation/opencensus-proto/blob/192f9920ad6a40dfc65e2e6732134d0d45d3fc89/src/opencensus/proto/trace/v1/trace.proto

c24t: I don't know about that. I think we're more likely to change the file in place than move it, and don't think we should link to an old version of the file from the readme.

bogdandrutu:
Already approved this change in a different PR. Please close this PR if that is a true statement.

c24t:
Looks like it was fixed in #224.

PR #230, opened 2019-01-21
fix tracestate link

discostu105:

PR #236, opened 2019-02-20
fix tracestate link

discostu105:


discostu105:
Re-did https://github.com/census-instrumentation/opencensus-specs/pull/230 again, now with CLA signed.

PR #235, opened 2019-02-20
Add a list of standard supported resources.

bogdandrutu:
The current list of standard resources includes only the resources we've already used in OC. We need to add support for more resources but this is just to sync what is defined in OC library with specs.

PR #238, opened 2019-02-27
Remove outdated resource specs.

songy23:
/cc @c24t


songy23:
See current Java implementation: https://github.com/census-instrumentation/opencensus-java/blob/master/contrib/resource_util/src/main/java/io/opencensus/contrib/resource/util.

songy23:
Apologies, just found this is duplicated to https://github.com/census-instrumentation/opencensus-specs/pull/235. Update this PR to only remove the stale specs.

songy23:
New specs on standard resources will be added in #235.

PR #254, opened 2019-03-28
Metrics: Add specs on Cumulative APIs

### LINKED ISSUE #241: Metrics: Add specs on Cumulative APIs.
Similar to #240 but the generated metrics will be cumulative and have start time. The value can only go up.

mayurkale22:
Fixes #241

### DIFF
--- a/metrics/Cumulative.md
+++ b/metrics/Cumulative.md
@@ -1,0 +1,51 @@
+# Cumulative Overview
+A `Cumulative` is used to record aggregated metrics that represents a single numerical value accumulated over a time interval. The value can only increase or be reset to zero on restart or reset the event. The recorded values are always >= 0. Typical examples of counters would be the number of elements added to a queue, total cpu usage, tasks completed, cache misses, etc.
+
+If you need to track or record a value that goes up and down, use a [Gauge](./Gauge.md) API.
+
+This document describes the key types and the overall behavior of API.
+
+## Cumulative API
+
+The value that is published for cumulative is an instantaneous measurement of an `int64` or `double` value. This API is useful when you want to manually increase and reset values as per service requirements.
+
+The following general operations MUST be provided by the API:
+* Defining a `name`, `description`, `unit`, `labelKeys`, `resource` and `constantLabels` which are fixed labels that always apply to a metric. This should give back the cumulative object to get or create time series, remove time series and clear all time series.
+	* `name`: a string describing the name of the metric, e.g. "task_completed" or "total_errors". Names MUST be unique within the library. It is recommended to use names compatible with the intended end usage.
+	* `description`: an optional string describing the metric, e.g."Total task completed". The default is set to "".
+	* `unit`: an optional string describing the unit used for the metric. Follows the format described by
+[Unified Code for Units of Measure](http://unitsofmeasure.org/ucum.html). The default set to "1".
+	* `labelKeys`: an optional list of the label keys to track different types of metric. The default is set to empty list.
+	* `constantLabels`: an optional map of label keys and label values. The default is set to empty map.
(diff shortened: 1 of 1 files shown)


metrics/Cumulative.md
```diff
@@ -0,0 +1,51 @@
+# Cumulative Overview
+A `Cumulative` is used to record aggregated metrics that represents a single numerical value accumulated over a time interval. The value can only increase or be reset to zero on restart or reset the event. The recorded values are always >= 0. Typical examples of counters would be the number of elements added to a queue, total cpu usage, tasks completed, cache misses, etc.
```
songy23: Typical examples of counters -> Typical examples of cumulative

mayurkale22:
> Question: how do we define and deal with start time with cumulative APIs? Do we set the start time when creating a cumulative and reset it to the time when `reset` is called?

In a time series, cumulative metric should have the same start time (set when creating a cumulative) and increasing end times, until an event resets then the start time should also be reset and we should use new start time for the following points.

I will update the specs with this info. Thanks

metrics/Cumulative.md
mayurkale22: Done

metrics/Cumulative.md
```diff
@@ -0,0 +1,54 @@
...
+	* `resource`: same as above resource.
+* Add a new time series with label values, an `object` and a `callback function`. The number of label values must be the same to that of the label keys.
+	* `labelValues`: the list of label values. The number of label values must be the same to that of the label keys.
+	* `object`: the state object from which the function derives a measurement.
+	* `function`: the callback function to be called.
+* Remove a single time series from the cumulative metric, if it is present.
+	* `labelValues`: the list of label values.
+* Clear all time series from the cumulative metric i.e. References to all previous point objects are invalid (not part of the metric).
+
+In a time series, cumulative metric should have the same start time (set when creating a cumulative) and increasing end times, until an event resets then the start time should also be reset and we should use new start time for the following points.
```
rghetia: What if the function registered returns a value that is lower than before (due to reset or some other reason) without application calling reset? Do we
1. reset the start time whenever we detect that new value is not higher than the old value. OR
2. continue to report old value.

I think we should do (1).

metrics/Cumulative.md
```diff
@@ -0,0 +1,54 @@
...
+	* `description`: an optional string describing the metric, e.g."Total task completed". The default is set to "".
+	* `unit`: an optional string describing the unit used for the metric. Follows the format described by
+[Unified Code for Units of Measure](http://unitsofmeasure.org/ucum.html). The default set to "1".
+	* `labelKeys`: an optional list of the label keys to track different types of metric. The default is set to empty list.
+	* `constantLabels`: an optional map of label keys and label values. The default is set to empty map.
+	* `resource`: an optional associated monitored resource information.
+* Add a new time series with label values, which returns a `Point` (which is part of the `TimeSeries`). Each point represents an instantaneous measurement of a cumulative value. Each Cumulative Metric has one or more time series for a single metric.
+	* `labelValues`: the list of label values. The number of label values must be the same to that of the label keys.
+* The `Point` class should provide functionalities to manually increment/reset values. Example: `inc()` bye one, `inc(long times)` a specific number of times at once, `reset()`.
```
rghetia: We need set() that does conditional sets the value if the old value is lower.
inc() and inc(long times) both may not be needed. For Go it would require two functions.

songy23: Chatted with @rghetia and @bogdandrutu offline, we agreed that a `Set` API for Cumulative is not needed. Users should be able to use the derived cumulative API for aggregated metrics rather than creating their own timer.

PR #275, opened 2022-12-09
gRPC Stats: Update definition of grpc.io/client/roundtrip_latency to reflect ground reality

yashykt:
cc @markdroth @ejona86 @dfawley

Background - Our original understanding was that the definition really was asking to collect the latency from the transport's perspective, but we noticed that the implementations (in gRPC C++, Java and Go) weren't really collecting it at the transport but more at the attempt-top. Given that that's the case, we figured we should update the definition to reflect the ground reality.


markdroth (approved):

punya:
@dashpole @jsuereth PTAL.

dashpole (approved):

